PR #20, opened 2020-05-17
Update Readme instructions for Pantheon.

adamhstewart:
I noticed when setting up a site for the first time, there are a few cases where the instructions could be a bit clearer.


amitaibu:
Lovely, thanks.

PR #35, opened 2020-06-15
Mkcert fix

AronNovak:
https://github.com/drud/ddev/issues/1848#issuecomment-553385735


amitaibu:
I assume present Aron was happy to see that past Aron already solved? :smile:

AronNovak:
Absolutely :)

PR #44, opened 2020-07-10
Basic perf settings

AronNovak:
we should not launch any project with anonymous caching turned off and CSS / JS aggregation disabled


amitaibu:
Good catch, thanks.

Issue #46, opened 2020-07-16
Check PHP syntax of custom code


mariano-dagostino:
PHPCS doesn't check php syntax, missing test coverage will not detect this either.

There is an interesting project called https://github.com/VKCOM/noverify that I've been trying to check syntax errors locally. Let's see if this can be automated.

A fatal error waiting to be discovered by someone:

![Screenshot-2020-07-16-17-06-23](https://user-images.githubusercontent.com/5313631/87717434-bf76cb00-c786-11ea-88ac-25f1d8c5cd5f.png)

PR #47, opened 2020-07-16
Try VKCOM/noverify to check PHP syntax of custom code

mariano-dagostino:
#46 

See https://github.com/Gizra/drupal-starter/runs/879224975  to have a sense of why I think this PR is useful.


amitaibu:
Please re-open if you think this is still needed

PR #49, opened 2020-08-02
Tailwind integration, Style guide and NodeViewBuilders

amitaibu:
1. Adding a style-guide module, so themes can be built in parallel to design. Visit - https://drupal-starter.ddev.site:4443/style-guide
image
1. Overriding `NodeViewBuilder` to control how nodes are rendered in different view modes. The benefit is that the entire node view is controlled from a single place -- removing the confusion of "where is this element coming from? Preprocess? Maybe a block, etc"


amitaibu:
@bboro ready for review when you have time

web/modules/custom/server_general/src/Controller/Homepage.php
```diff
@@ -0,0 +1,46 @@
...
+  }
+
+  /**
+   * Build the hero header.
+   *
+   * @return array
+   *   Render array.
+   */
+  protected function buildMainContent() {
+    $element = ['#markup' => $this->t('Add your Homepage elements in \Drupal\server_general\Controller\Homepage')];
```
bboro, line 31: :+1:

web/modules/custom/server_general/src/NodeViewBuilderAbstract.php
```diff
@@ -0,0 +1,242 @@
...
+
+namespace Drupal\server_general;
+
+use Drupal\Core\Block\BlockManagerInterface;
+use Drupal\Core\Entity\EntityTypeManager;
+use Drupal\Core\Session\AccountInterface;
+use Drupal\node\NodeInterface;
+
+/**
+ * Class NodeViewBuilderPerBundleAbstract.
```
bboro, line 11:
```suggestion
 * Class NodeViewBuilderAbstract.
```

web/modules/custom/server_general/src/NodeViewBuilderAbstract.php
```diff
@@ -0,0 +1,242 @@
...
+    return $build;
+  }
+
+  /**
+   * Get common elements for the view modes.
+   *
+   * @param \Drupal\node\NodeInterface $entity
+   *   The entity.
+   *
+   * @return array[]
```
bboro, line 85:
```suggestion
   * @return array
```

or 

```suggestion
   * @return mixed[]
```

(id is an int I think, others are string)

web/modules/custom/server_general/src/NodeViewBuilderAbstract.php
```diff
@@ -0,0 +1,242 @@
...
+
+  /**
+   * Build an image referenced in the given entity's given field name.
+   *
+   * @param \Drupal\node\NodeInterface $entity
+   *   The entity.
+   * @param string $field_name
+   *   Optional; The field name. Defaults to "field_image".
+   *
+   * @return array|string
```
bboro, line 178:
```suggestion
   * @return array
```

or

```suggestion
   * @return string[]
```

web/modules/custom/server_general/src/NodeViewBuilderArticle.php
```diff
@@ -0,0 +1,58 @@
...
+namespace Drupal\server_general;
+
+use Drupal\node\NodeInterface;
+
+/**
+ * Class NodeViewBuilderArticle.
+ */
+class NodeViewBuilderArticle extends NodeViewBuilderAbstract {
+
+  use TagBuilderTrait;
```
bboro, line 12: I think we should put this in the Abstract class since the abstract class also has `buildTags` method. Unless we're trying to reduce the amount of unneeded code in other node type builders? If so perhaps we should move buildTags into the trait instead?

web/modules/custom/server_general/src/TagBuilderTrait.php
```diff
@@ -0,0 +1,37 @@
...
+   * Get a tag.
+   *
+   * @param \Drupal\taxonomy\TermInterface $term
+   *   The term to render.
+   *
+   * @return array
+   *   The renderable array.
+   */
+  public function buildTag(TermInterface $term) {
+    $classes = 'mr-1 text-ms px-3 py-1 my-1 text-center leading-normal rounded-large border-2 border-purple-primary hover:text-blue-900 hover:border-blue-900 rounded-md text-purple-primary h-8 overflow-hidden';
```
bboro, line 25: This one feels like an outlier. Maybe worth a `server_theme_tag` theme. That will also help with `purge-css` to detect the classes

web/modules/custom/server_style_guide/src/Controller/StyleGuideController.php
```diff
@@ -0,0 +1,320 @@
...
+      $this->getMockedTag('bare klingon'),
+      $this->getMockedTag('Dogma doesn’t balanced understand'),
+      $this->getMockedTag('The plank hails with courage'),
+      $this->getMockedTag('burn the freighter until it rises'),
+    ];
+
+    $single_card_simple = [
+      '#theme' => 'server_theme_card__simple',
+      '#image' => $card_image,
+      '#title' => 'The source has extend, but not everyone fears it.',
```
bboro, line 93:
```suggestion
      '#title' => $this->t('The source has extend, but not everyone fears it.'),
```

Do we wanna bother translating this gibberish? There are a few more missed ones below. I won't highlight them all.

web/modules/custom/server_style_guide/templates/server-style-guide-header.html.twig
```diff
@@ -0,0 +1,11 @@
+<div class="flex flex-row items-baseline">
+  <div class="mb-4 mt-8">
+    <a class="text-blue-500 text-3xl font-bold" href="#{{ unique_id }}" id="{{ unique_id }}">{{ title }}</a>
+  </div>
+
+  {% if link %}
+    <div class="ml-4">
+      <a class="text-sm text-blue-500" href="{{ link }}" target="_blank">See design</a>
```
bboro, line 8:
```suggestion
      <a class="text-sm text-blue-500" href="{{ link }}" target="_blank">{{ 'See design'|t }}</a>
```

web/themes/custom/server_theme/templates/page.html.twig
```diff
@@ -42,13 +42,14 @@
  * @see html.html.twig
  */
 #}
-<div class="layout-container">
+<div class="layout-container text-black flex-grow">
```
bboro, line 45: Will the `flex-grow` do anything here? The element itself is not `display: flex` without the `flex` class, and the parent is a `<body>` right?

web/themes/custom/server_theme/templates/server-theme-button.html.twig
```diff
@@ -0,0 +1,72 @@
...
+    {% set button_attributes = button_attributes.setAttribute('onclick', onclick) %}
+  {% endif %}
+  {% if target %}
+    {% set button_attributes = button_attributes.setAttribute('target', target) %}
+  {% endif %}
+  {% if icon_only %}
+    {% set button_attributes = button_attributes.setAttribute('title', label) %}
+  {% endif %}
+{% else %}
+  {% set button_attributes = button_attributes.addClass(classes|join(' ')) %}
```
bboro, line 60:
```suggestion
  {% set button_attributes = button_attributes.addClass(classes) %}
```
We can add an array of classes, my bad for originally doing it with a `join` on portal :D

web/modules/custom/server_style_guide/src/Controller/StyleGuideController.php
amitaibu: Neh, I think we can avoid the `$this->t()`

web/modules/custom/server_general/src/TagBuilderTrait.php
amitaibu: Thanks, I'll move. btw, `purge-css` is now scanning those files as-well

bboro: Neat!

web/themes/custom/server_theme/templates/page.html.twig
amitaibu: Good catch

PR #51, opened 2020-08-03
Port RoboFile improvements from client project 

AronNovak:
#50

### DIFF
--- a/RoboFile.php
+++ b/RoboFile.php
@@ -180,19 +180,19 @@ class RoboFile extends Tasks {
   /**
    * Deploy to Pantheon.
    *
-   * @param string $branchName
+   * @param string $branch_name
    *   The branch name to commit to. Default to master.
    *
    * @throws \Exception
    */
-  public function deployPantheon($branchName = 'master') {
+  public function deployPantheon($branch_name = 'master') {
     if (empty(self::PANTHEON_NAME)) {
       throw new Exception('You need to fill the "PANTHEON_NAME" const in the Robo file. so it will know what is the name of your site.');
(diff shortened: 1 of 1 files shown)


RoboFile.php
```diff
@@ -339,10 +351,10 @@ public function phpcs() {
     $directories = [
       'modules/custom',
       'themes/custom',
-      'profiles/custom'
+      'profiles/custom',
     ];
 
-    $errorCode = null;
+    $errorCode = NULL;
```
amitaibu, line 357:
```suggestion
    $error_code = NULL;
```

RoboFile.php
```diff
@@ -238,7 +243,7 @@ public function deployPantheon($branchName = 'master') {
       'travis-key',
     ];
 
-    $rsyncExcludeString = '--exclude=' . join(' --exclude=', $rsyncExclude);
+    $rsyncExcludeString = '--exclude=' . implode(' --exclude=', $rsyncExclude);
```
amitaibu, line 246: change to snake_case

AronNovak:
@amitaibu That would require more work, I could not return back to that, introducing that is not obvious, as it's not part of Terminus itself. I guess it's better to keep Drupal Starter up-to-date early, as this race condition leads to build failures.

amitaibu:
> I guess it's better to keep Drupal Starter up-to-date early, as this race condition leads to build failures.

Ok, sounds good.

AronNovak:
@amitaibu There were lots of camelCase occurrences, I eliminated them all hopefully now.

PR #52, opened 2020-08-03
More helpful auto-depoy-config output

AronNovak:
#50

### DIFF
--- a/RoboFile.php
+++ b/RoboFile.php
@@ -456,6 +456,8 @@ class RoboFile extends Tasks {
     $this->say("The project was prepared for the automatic deployment to Pantheon");
     $this->say("Review the changes and make a commit from the added files.");
     $this->say("Add the SSH key to the Pantheon account: https://pantheon.io/docs/ssh-keys .");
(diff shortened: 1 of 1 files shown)


RoboFile.php
```diff
@@ -456,6 +456,8 @@ public function deployConfigAutodeploy(string $token, string $project_name, $git
     $this->say("The project was prepared for the automatic deployment to Pantheon");
     $this->say("Review the changes and make a commit from the added files.");
     $this->say("Add the SSH key to the Pantheon account: https://pantheon.io/docs/ssh-keys .");
+    $this->say("Add the SSH key to the GitHub project as a deploy key: https://docs.github.com/en/developers/overview/managing-deploy-keys .");
+    $this->say("Convert the project to nested docroot: https://pantheon.io/docs/nested-docroot .");
```
amitaibu, line 460: We already confirm this here - https://github.com/Gizra/drupal-starter/blob/1ddaddad08dcbb2c74734d8319a3125561732621/RoboFile.php#L225-L227

AronNovak: But we do not confirm if the repository was already converted, it's not just about switching that boolean flag, my idea was that we link the documentation about how to do it properly.

PR #55, opened 2020-08-09
Add Entity View Build plugins

amitaibu:
Instead of services to hijack an entity view, we'll use Plugins.

Outcome remains the same

image

However, we have less wiring to do every-time we'd like to add another Plugin. Just write the class - and we'd send it to the plugin if it exists. Otherwise, we'll return the original build.

/cc @bboro


web/modules/custom/server_general/src/EntityViewBuilder/EntityViewBuilderPluginInterface.php
```diff
@@ -0,0 +1,31 @@
...
+   *   The entity to show.
+   * @param string $view_mode
+   *   The view mode. Defaults to "full".
+   * @param string $langcode
+   *   Optional; The language code.
+   *
+   * @return array
+   *   The new render array.
+   */
+  public function build(array $build, EntityInterface $entity, $view_mode = 'full', $langcode = NULL);
```
amitaibu, line 29: I'll change `build` to `view` since we wire it from the `EntityViewBuilder::view`

PR #62, opened 2020-09-06
Use Tailwind's purge whitelist instead of the purgecss.html.twig workaround file

ordavidil:
https://github.com/Gizra/un-tp/pull/172#issuecomment-687713830


amitaibu:
Thanks.

PR #66, opened 2020-09-07
Freeze coder module inside Travis

AronNovak:


ci-scripts/install_coder.sh
```diff
@@ -8,5 +8,5 @@ set -e
 # ---------------------------------------------------------------------------- #
 
 cd "$TRAVIS_BUILD_DIR"
-COMPOSER_MEMORY_LIMIT=-1 composer global require drupal/coder:^8.3.1
+COMPOSER_MEMORY_LIMIT=-1 composer global require drupal/coder:8.3.9
```
amitaibu, line 11: I think it's nice we get the most up to date version. You disagree?

AronNovak: @amitaibu right, for our starter kit, we should upgrade this.

AronNovak: Especially that no code change is needed to pass :)

ci-scripts/install_coder.sh
```diff
@@ -8,5 +8,5 @@ set -e
 # ---------------------------------------------------------------------------- #
 
 cd "$TRAVIS_BUILD_DIR"
-COMPOSER_MEMORY_LIMIT=-1 composer global require drupal/coder:^8.3.1
+COMPOSER_MEMORY_LIMIT=-1 composer global require drupal/coder:8.3.10
```
amitaibu, line 11: Why not?

```suggestion
COMPOSER_MEMORY_LIMIT=-1 composer global require drupal/coder:^8.3.10
```

AronNovak: As it means it can fail accidentally. Generally we need to have exact, freezed versions in Travis, so we avoid suprising failures on release days of the packages we use.

AronNovak: `^8.3.10` means that it can be 8.3.15 too.

amitaibu: Thing is that without `^` we won't notice updates in coder module. So I think that if a project has failure they can't fix quickly, then on that project they should freeze it.

With that said - If you still feel the right version is without `^`, then ho ahead and merge :)

AronNovak: Alright, let's keep it unfreezed in the kit, it indeed has some benefits.

PR #68, opened 2020-09-16
Add a robo command to reinstall an environment on Pantheon.

ordavidil:
Reinstalling an environment on Pantheon requires running multiple commands.
Instead of running them manually each time, it's better to have a script to do it.

### DIFF
--- a/RoboFile.php
+++ b/RoboFile.php
@@ -338,10 +338,8 @@ class RoboFile extends Tasks {
   /**
    * Reinstall the site on specific env on Pantheon.
    *
-   * Running this command via `ddev` will require terminus login inside ddev.
-   * Instead, it's possible to run the command locally, where usually already
-   * authenticated to `terminus`:
-   * - `./vendor/consolidation/robo/robo reinstall:env`
+   * Running this command via `ddev` will require terminus login inside ddev:
+   * `ddev auth ssh`
    *
    * @param string $env
    *   The environment to reinstall (default='qa').
--- a/RoboFile.php
+++ b/RoboFile.php
@@ -336,6 +336,58 @@ class RoboFile extends Tasks {
   }
 
   /**
+   * Reinstall the site on specific env on Pantheon.
+   *
+   * Running this command via `ddev` will require terminus login inside ddev.
+   * Instead, it's possible to run the command locally, where usually already
+   * authenticated to `terminus`:
+   * - `./vendor/consolidation/robo/robo reinstall:env`
+   *
+   * @param string $env
+   *   The environment to reinstall (default='qa').
+   *
+   * @throws \Robo\Exception\TaskException
+   */
+  public function reinstallEnv(string $env = 'qa') {
+    $forbidden_envs = [
+      'live',
+    ];
+    if (in_array($env, $forbidden_envs)) {
+      throw new Exception("Reinstalling the site on `{$env}` environment is forbidden.");
+    }
+
+    $pantheon_name = self::PANTHEON_NAME;
+    $pantheon_terminus_environment = $pantheon_name . '.' . $env;
+
+    // This set of commands should work, so expecting no failures.
+    // (tend to invoke the same flow as DDEV's `config.local.yaml`.
+    $task = $this
+      ->taskExecStack()
+      ->stopOnFail();
+
+    $result = $task
+      ->exec("terminus remote:drush $pantheon_terminus_environment -- si server -y --existing-config")
+      ->exec("terminus remote:drush $pantheon_terminus_environment -- en server_migrate -y")
+      ->exec("terminus remote:drush $pantheon_terminus_environment -- migrate:import --group=server")
+      ->exec("terminus remote:drush $pantheon_terminus_environment -- pm:uninstall server_migrate")
+      ->exec("terminus remote:drush $pantheon_terminus_environment -- uli")
+      ->run()
+      ->getExitCode();
+
+    // For these environments, set the `admin` user's password to `1234`.
+    $envs_to_set_admin_simple_password = [
+      'qa'
+    ];
+    if (in_array($env, $envs_to_set_admin_simple_password)) {
+      $task->exec("terminus remote:drush $pantheon_terminus_environment -- user:password admin 1234")->run();
+    }
+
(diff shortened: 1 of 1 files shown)


RoboFile.php
```diff
@@ -335,6 +335,58 @@ public function deployPantheonSync(string $env = 'test', bool $do_deploy = TRUE)
     }
   }
 
+  /**
+   * Reinstall the site on specific env on Pantheon.
+   *
+   * Running this command via `ddev` will require terminus login inside ddev.
+   * Instead, it's possible to run the command locally, where usually already
+   * authenticated to `terminus`:
+   * - `./vendor/consolidation/robo/robo reinstall:env`
```
amitaibu, line 344: Yeah, that's a good thing :) In fact, all you need to do is https://github.com/Gizra/drupal-starter#executing `ddev auth ssh`

RoboFile.php
```diff
@@ -335,6 +335,58 @@ public function deployPantheonSync(string $env = 'test', bool $do_deploy = TRUE)
...
+   *
+   * @param string $env
+   *   The environment to reinstall (default='qa').
+   *
+   * @throws \Robo\Exception\TaskException
+   */
+  public function reinstallEnv(string $env = 'qa') {
+    $forbidden_envs = [
```
amitaibu, line 352: :+1:

RoboFile.php
```diff
@@ -335,6 +335,58 @@ public function deployPantheonSync(string $env = 'test', bool $do_deploy = TRUE)
...
+      ->exec("terminus remote:drush $pantheon_terminus_environment -- si server -y --existing-config")
+      ->exec("terminus remote:drush $pantheon_terminus_environment -- en server_migrate -y")
+      ->exec("terminus remote:drush $pantheon_terminus_environment -- migrate:import --group=server")
+      ->exec("terminus remote:drush $pantheon_terminus_environment -- pm:uninstall server_migrate")
+      ->exec("terminus remote:drush $pantheon_terminus_environment -- uli")
+      ->run()
+      ->getExitCode();
+
+    // For these environments, set the `admin` user's password to `1234`.
+    $envs_to_set_admin_simple_password = [
```
amitaibu, line 378: :+1:

amitaibu (commented):
Thanks, looks good. Couple of comments + letting @AronNovak also do a review

RoboFile.php
amitaibu: A good thing == running from inside DDEV. So one doesn't even need to have terminus locally for this to run

RoboFile.php
```diff
@@ -335,6 +335,58 @@ public function deployPantheonSync(string $env = 'test', bool $do_deploy = TRUE)
...
+      'qa'
+    ];
+    if (in_array($env, $envs_to_set_admin_simple_password)) {
+      $task->exec("terminus remote:drush $pantheon_terminus_environment -- user:password admin 1234")->run();
```
AronNovak, line 382: My only concern is from the cybersec point of view. This is kind of scary that we document it this way. We might think about the strategy in general how we set passwords for admin 1 . Even non-live client sites are full of very sensitive data. We have one more layer of security. But if it's turned off for any reason, temporarily for a demo, or if someone forgets a setup step... . So one idea that comes into my mind that we keep the random-generated password, but the email is set to a special address. Or let a Slack robot / Productivity provide you a one-time link, where personal authentication is already solved.

amitaibu: I see your point, but sounds a bit too complex IMO - not only the Slack integration, but also the fact that QA/ PMs will need another place to look for new passwords.

How about:

1. have a switch to decide if we should simplify passwords at all.

```diff
- if (in_array($env, $envs_to_set_admin_simple_password)) {
+ if (!empty($envs_to_set_admin_simple_password) &&in_array($env, $envs_to_set_admin_simple_password)) {
```

2. When scaffolding a site auto set it to be .httacess protected?

AronNovak: Good point, I'll think about how to make our first layer of protection fully automated

amitaibu: `terminus lock:enable <site>:<env> -- user password` - https://pantheon.io/docs/security

AronNovak: Yep., that would work. I guess the best would be to create a Jenkins job that ensures that across all the sites, we have a coherent htaccess for all the non-live environments every day. This way, if we need to change the password for any reason, we can do it trivially from one place. And in addition, of course, the RoboFile could include this too.

amitaibu: > I guess the best would be to create a Jenkins job that ensures that across all the sites

Not necessarily - some you'd want to keep open, since for example you could have REST calls into it, and having those with .htaccess has proven to be sometimes hard. Also, probably out of scope for this issue.

IMO provisioning the env locked (upon install) is "good enough"

AronNovak: Ok, I am convinced, let"s add the flag here, if we need to simplify the password or not.

RoboFile.php
AronNovak: ^^ @ordavidil are you going to cover this?

(I closed the other thread, generally the provisioning of a new site is not automated, that's the high-level issue, like creating GH project, creating Pantheon project + its environments)

ordavidil: you can take over is you wish

amitaibu:
Thanks

RoboFile.php
```diff
@@ -335,6 +335,56 @@ public function deployPantheonSync(string $env = 'test', bool $do_deploy = TRUE)
...
+   *
+   * Running this command via `ddev` will require terminus login inside ddev:
+   * `ddev auth ssh`
+   *
+   * @param string $env
+   *   The environment to reinstall (default='qa').
+   *
+   * @throws \Robo\Exception\TaskException
+   */
+  public function reinstallEnv(string $env = 'qa') {
```
amitaibu, line 349: @AronNovak I think we can have a follow up here to:

1. Update the README to use this command
1. Maybe change the name to `pantheonInstallEnv`?

amitaibu: https://github.com/Gizra/drupal-starter#install-the-site-with-the-profile

PR #69, opened 2020-09-23
Compress the SVGs while compiling the theme.

ordavidil:


README.md
```diff
@@ -57,6 +57,11 @@ The directory structure:
 
 For theme development, it's advisable to entirely turn off caching: https://www.drupal.org/node/2598914
 
+#### Compressing SVG files
```
amitaibu, line 60: I don't think we need to mention this specifically in the README, since it's part of the `theme:compile`

RoboFile.php
```diff
@@ -335,6 +338,28 @@ public function deployPantheonSync(string $env = 'test', bool $do_deploy = TRUE)
     }
   }
 
+  /**
+   * Compress SVG files in specific directories.
```
amitaibu, line 342: Mention this is called as part of the theme:compile

amitaibu:
Thanks

PR #87, opened 2020-11-19
Fix svg compress error and working directory

amitaibu:
* PR prevents this error when no SVG file exists

image

* We perform the optimization on the `dist` directory instead of directly on the `src`

### DIFF
--- a/RoboFile.php
+++ b/RoboFile.php
@@ -109,7 +109,7 @@ class RoboFile extends Tasks {
         ->run();
 
       // Compress all SVGs.
-      $this->themeCompressSvg();
+      $this->themeSvgCompress();
     }
 
     $this->_exec('drush cache:rebuild');
@@ -349,7 +349,7 @@ class RoboFile extends Tasks {
    * This function is being called as part of `theme:compile`.
    * @see compileTheme_()
    */
-  public function themeCompressSvg() {
+  public function themeSvgCompress() {
     $directories = [
       './dist/images',
     ];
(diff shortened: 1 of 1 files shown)


RoboFile.php
```diff
@@ -343,20 +344,28 @@ public function deployPantheonSync(string $env = 'test', bool $do_deploy = TRUE)
...
 
     foreach ($directories as $directory) {
-      $result = $this->_exec("cd web/themes/custom/server_theme && ./node_modules/svgo/bin/svgo $directory");
+      // Check if SVG files exists in this directory.
+      $finder = new Finder();
+      $finder->in('web/themes/custom/server_theme/' . $directory);
+      if (!$finder->hasResults()) {
+        // No SVG files.
+        continue;
+      }
```
ordavidil, line 366: These doesn't seems to specifically search SVG files, so the comments are confusing.

ordavidil (commented):
>  * We perform the optimization on the `dist` directory instead of directly on the `src`

I think that part of the point of compressing the `src` is to optimize the repo size as well (at least from the recent projects I've been working on).

RoboFile.php
amitaibu: Indeed :+1:  - added correct logic.

I've also renamed function so command would be `ddev robo theme:svg-compress`

amitaibu:
Seems `svgo` got some files corrupted in one of our projects, and replaced with `svgcleaner`. So I'll replace it here as-well

/cc @anvmn

amitaibu:
Actually, lets get it in a follow up.

PR #88, opened 2020-11-20
Tailwind 2 with PostCSS

amitaibu:
PR upgrades to Tailwind 2.

Also, following https://tailwindcss.com/docs/using-with-preprocessors#using-post-css-as-your-preprocessor,  Instead of compiling with SASS itself along with PostCSS, we're using `postcss-sass`, and a bunch of other postcss plugins, which removes dependency in any SASS libs or parcelJs


amitaibu:
Example showing it's still working:

image

README.md
```diff
@@ -41,15 +41,15 @@ then execute the following, and re-try installation steps.
 
 By default, `ddev restart` compiles the theme using Robo.
 
-On the local development environment, execute:
+On the local development environment, which is using TailWind, execute:
 ```bash
-ddev robo theme:watch-debug
+ddev robo theme:complie-debug
```
bboro, line 46:
```suggestion
ddev robo theme:compile-debug
```

web/themes/custom/server_theme/src/scss/style.scss
```diff
@@ -1,6 +1,10 @@
-@tailwind base;
-@tailwind components;
-@tailwind utilities;
+@import "tailwindcss/base";
+@import "tailwindcss/components";
+
+@import "components/typography.scss";
+@import "components/list.scss";
```
bboro, line 5: if we rename the files to be sass "partials" (starting with underscore: `_typography.scss`) we can import them without the file extension:

```scss
@import "components/typography";
```

bboro (commented):
Couple tiny things. Looks great otherwise!

web/themes/custom/server_theme/src/scss/style.scss
amitaibu: I think it no longer works because we're not really use SASS anymore. That is,  `postcss-sass` is responsible for the parsing, and postcss-import is responsible for the `@import`, which is now strict like CSS, and requires the full name

bboro: Oh I see :+1:

bboro (approved):

PR #95, opened 2020-11-26
Release notes beauty fix

AronNovak:
## Status
image

### DIFF
--- a/RoboFile.php
+++ b/RoboFile.php
@@ -919,7 +919,7 @@ END;
       preg_match_all('!from Gizra/([0-9]+)!', $line, $issue_matches);
 
       if (isset($issue_matches[1][0])) {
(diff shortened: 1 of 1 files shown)


RoboFile.php
```diff
@@ -919,7 +919,7 @@ public function generateReleaseNotes($tag = NULL) {
       preg_match_all('!from Gizra/([0-9]+)!', $line, $issue_matches);
```
amitaibu, line 919: Are we able to make this non-Gizra specific?

amitaibu (approved):

RoboFile.php
AronNovak: Sure, added an expression instead of Gizra

AronNovak:
image
issue detection still works, where possible

Issue #98, opened 2020-12-01
Kernel Tests are not working


mariano-dagostino:
```
There was 1 error:

1) Drupal\Tests\server_general\Kernel\ServerGeneralTest::testDummy
PHPUnit\Framework\Exception: PHP Fatal error:  Uncaught PHPUnit\Framework\Exception: Could not read "phpunit.xml.dist".
```

mariano-dagostino:
@amitaibu @AronNovak any clue how to make this work? https://travis-ci.com/github/Gizra/drupal-starter/jobs/451446472#L1295

mariano-dagostino:
I'm asking this because we could use drupal-starter to easily run kernel and browser tests when doing some contrib code.

amitaibu:
hmm, I'm was able to run Kernel tests when developing contrib modules.

So I don't know about the travis failure, but what happens when you try to run your contrib module using `ddev phpunit --filter=ClassName`?

mariano-dagostino:
@amitaibu same error:

```bash
> ddev phpunit --filter=ServerGeneralTest
PHPUnit 7.5.20 by Sebastian Bergmann and contributors.

E                                                                   1 / 1 (100%)

Time: 400 ms, Memory: 4.00 MB

There was 1 error:

1) Drupal\Tests\server_general\Kernel\ServerGeneralTest::testDummy
PHPUnit\Framework\Exception: PHP Fatal error:  Uncaught PHPUnit\Framework\Exception: Could not read "phpunit.xml.dist".

  thrown in /var/www/html/vendor/phpunit/phpunit/src/Util/Configuration.php on line 196
Fatal error: Uncaught PHPUnit\Framework\Exception: Could not read "phpunit.xml.dist".
```

amitaibu:
Are you most up to date `ddev composer install`?

mariano-dagostino:
After configuring properly phpunit.xml from web/core now I'm getting  

```
1) Drupal\Tests\server_general\Kernel\ServerGeneralTest::testDummy
Error: Class 'org\bovigo\vfs\vfsStream' not found
```

Seems related to this https://www.drupal.org/node/2889302

mariano-dagostino:
@amitaibu I tried this again today with a newer version of drupal and now is working out of the box.

PR #99, opened 2020-12-01
Kernel tests are not working

mariano-dagostino:
#98

### DIFF
--- a/.ddev/commands/web/phpunit
+++ b/.ddev/commands/web/phpunit
@@ -4,4 +4,4 @@
 ## Usage: phpunit
 ## Example: "ddev phpunit" or "ddev phpunit web/modules/custom/server_general".
 
-cd .. && ./vendor/bin/phpunit -c phpunit.xml.dist  $@
+cd .. && ./vendor/bin/phpunit $@
--- a/composer.json
+++ b/composer.json
@@ -37,6 +37,7 @@
         "dealerdirect/phpcodesniffer-composer-installer": "^0.7.0",
         "drupal/coder": "^8.3.1",
         "henrikbjorn/lurker": "1.*",
+        "mikey179/vfsstream": "^1.6",
         "natxet/cssmin": "^3.0",
         "patchwork/jsqueeze": "^2.0",
         "phpunit/phpunit": "^7",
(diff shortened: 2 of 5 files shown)


amitaibu:
Is there a reason for Kernel tests on top of `ExistingSiteTest`?

amitaibu:
Ignore above comment, continuing discussion in #98

.ddev/commands/web/phpunit
```diff
@@ -4,4 +4,4 @@
 ## Usage: phpunit
 ## Example: "ddev phpunit" or "ddev phpunit web/modules/custom/server_general".
 
-cd .. && ./vendor/bin/phpunit -c phpunit.xml.dist  $@
+cd .. && ./vendor/bin/phpunit $@
```
mariano-dagostino, line 7: @amitaibu Not any clue why this is required, but passing a config file make the runner fail:

```
There was 1 error:

1) Drupal\Tests\server_general\Kernel\ServerGeneralTest::testDummy
PHPUnit\Framework\Exception: PHP Fatal error:  Uncaught PHPUnit\Framework\Exception: Could not read "phpunit.xml.dist".
```

composer.json
```diff
@@ -37,6 +37,7 @@
         "dealerdirect/phpcodesniffer-composer-installer": "^0.7.0",
         "drupal/coder": "^8.3.1",
         "henrikbjorn/lurker": "1.*",
+        "mikey179/vfsstream": "^1.6",
```
mariano-dagostino, line 40: @amitaibu Even if this is a dependency of drush, I'm not really sure why is not added to the list of dependencies. Adding without this we get:

```
There was 1 error:

1) Drupal\Tests\server_general\Kernel\ServerGeneralTest::testDummy
Error: Class 'org\bovigo\vfs\vfsStream' not found

```

.ddev/commands/web/phpunit
amitaibu: @mariano-dagostino can you please confirm the file is indeed there. That is, try `ddev ssh`, and try to run it manually, and see if you are able to find the problem

amitaibu: ok, please adapt PR accordingly - so it includes explicit call to `phpunit.xml.dist`

amitaibu:
Needs a re-roll

mariano-dagostino:
It seems this now works for D9 OOTB #98

PR #100, opened 2020-12-01
Use standard bootstrap for phpunit instead of the 'fast' one.

ordavidil:


amitaibu:
We've decided to keep the `fast`, and if there are class missing errors, they should be added to the custom `phpunit.xml.dist` file

PR #101, opened 2020-12-03
Regenerate config with ddev 1.16.2

bboro:
Nothing meaningful changed but in the spirit of keeping up to date.


amitaibu:
Thanks

PR #102, opened 2020-12-04
DDEV auto-phpcs command

AronNovak:
auto-phpcs-2020-12-04_11 13 08

### DIFF
--- a/.ddev/commands/web/auto-phpcs
+++ b/.ddev/commands/web/auto-phpcs
@@ -1,0 +1,10 @@
+#!/bin/bash
+
+## Description: Automatically corrects coding standard issues.
+## Usage: auto-phpcs
+## Example: "ddev auto-phpcs".
+
+cd ..
+while inotifywait -r -e modify web/modules/custom; do
+  robo phpcs
+done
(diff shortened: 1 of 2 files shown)


AronNovak:
@amitaibu Similarly to elm-format, certain things could happen on-the-fly. The only drawback is that it can generate conflicts between the IDE version and the disk version of the files, but Phpstorm seems to handle it nicely.

.ddev/commands/web/auto-phpcs
```diff
@@ -0,0 +1,10 @@
+#!/bin/bash
+
+## Description: Automatically corrects coding standard issues.
+## Usage: auto-phpcs
+## Example: "ddev auto-phpcs".
+
+cd ..
+while inotifywait -r -e modify web/modules/custom; do
```
amitaibu, line 8: I'm not sure we'd want a robo command for that. AFAIK you could have for example PHPStorm do that for you

AronNovak: Right, wiring into the IDE is smoother, not such a great idea in the end.

PR #103, opened 2020-12-05
Generic theme

adamhstewart:
A theme which is a bit more generic with 
- no color
- no copyrighted fonts
- a bit more appealing header/titles

image


web/themes/custom/server_theme/templates/purgecss.html.twig
```diff
@@ -0,0 +1,33 @@
...
+we are using, so they will not be purged by TailWind's purge CSS when we deoply.
+
+The way CSS purge works is by going over the twig/theme functions and look for
+the class names. If it will see in twig `bg-{{color}}` it will not pick it up,
+as it doesn't know to what it will comiple.
+
+Below are the custom colors we defined in tailwind.config.js
+#}
+
+black
```
amitaibu, line 11: This should be moved to tailwind.config.js under the existing `whitelist`

web/themes/custom/server_theme/templates/purgecss.html.twig
```diff
@@ -0,0 +1,33 @@
...
+gray-3
+gray-6
+gray-9
+gray-c
+white
+
+grid-cols-cards
```
amitaibu, line 18: why is this needed, isn't it auto-picked up from the twig file?

web/themes/custom/server_theme/tailwind.config.js
```diff
@@ -4,8 +4,12 @@ module.exports = {
   theme: {
     extend: {
       colors: {
-        'purple-primary': '#664493',
-        'turquoise': '#03B3AF',
+        'black': '#000000',
+        'gray-3': '#333333',
```
amitaibu, line 8: since we're removing design, maybe lets use one of the existing grays? https://tailwindcss.com/docs/customizing-colors

web/themes/custom/server_theme/tailwind.config.js
```diff
@@ -4,8 +4,12 @@ module.exports = {
...
+        'gray-6': '#666666',
+        'gray-9': '#999999',
+        'gray-c': '#cccccc',
+        'white' : '#ffffff',
```
amitaibu, line 12: we already have `text-white` and `text-black` by default - https://tailwindcss.com/docs/text-color

web/themes/custom/server_theme/server_theme.libraries.yml
```diff
@@ -10,6 +10,16 @@ global-scripts:
   dependencies:
     - core/jquery
 
+librefranklin:
```
amitaibu, line 13: Maybe lets replace with https://fonts.google.com/specimen/Roboto , which seems to be more popular?

web/themes/custom/server_theme/src/scss/components/typography.scss
```diff
@@ -1,17 +1,22 @@
+body {
+  font-family: 'Open Sans', sans-serif;
```
amitaibu, line 2: This should be on the twig file

web/themes/custom/server_theme/src/scss/components/typography.scss
```diff
@@ -1,17 +1,22 @@
...
+}
+
 h1, h2, h3, h4 {
-  @apply leading-none text-turquoise;
+  font-family: 'Libre Franklin', sans-serif;
```
amitaibu, line 6: This should be changed in tailwind.config.js - we already have https://github.com/Gizra/drupal-starter/blob/bebe2b00a00a228ad0d81b38871def43ae07a7ab/web/themes/custom/server_theme/tailwind.config.js#L29-L30

Rule of thumb is that we always try to stick to `@apply`, thus making sure we have a better consistency, and no CSS rules "in the wild"

amitaibu (commented):
Thanks for tackling this! I've added a few comments.

web/themes/custom/server_theme/src/scss/components/typography.scss
adamhstewart: or `@apply font-body`?

web/themes/custom/server_theme/templates/purgecss.html.twig
adamhstewart: @amitaibu is the purgecss file an outdated method? - if so I'll remove it from other projects too.

adamhstewart:
@amitaibu I think I hit everything - ready for round 2 at your convenience.

web/themes/custom/server_theme/src/scss/components/typography.scss
amitaibu: I prefer to add a CSS rule only when it's too complicated to add it to twig. What's nice about TW is that by reading a twig file, I know how it will behave

web/themes/custom/server_theme/templates/purgecss.html.twig
amitaibu: > is the purgecss file an outdated method?

Yes. `whitelist` is the right way

web/themes/custom/server_theme/src/scss/components/typography.scss
adamhstewart: Ah, I see on the body tag in particular....Headers, we still want to declare in CSS so it applies over all uses (like sizes)?

adamhstewart: or better even? https://tailwindcss.com/docs/adding-base-styles#using-a-plugin

web/themes/custom/server_theme/tailwind.config.js
```diff
@@ -1,12 +1,14 @@
 const plugin = require('tailwindcss/plugin');
+const colors = require('tailwindcss/colors')
 
 module.exports = {
   theme: {
+    colors: {
```
amitaibu, line 6: I think we can remove this part entirely, and have all TW's default colors loaded by default

web/themes/custom/server_theme/src/scss/components/typography.scss
amitaibu: > or better even? https://tailwindcss.com/docs/adding-base-styles#using-a-plugin

In the case of h1-h5 I'd keep it in the scss file.

With that said, after reading the Refactoring UI book, I understand that giving h1 the biggest size, and h5 the lowest is most probably not the right way.

adamhstewart: > I understand that giving h1 the biggest size, and h5 the lowest is most probably not the right way.

never was... Every homepage ever is an example of this. IMO sizes should live in the templates.

web/themes/custom/server_theme/server_theme.theme
```diff
@@ -141,3 +148,12 @@ function server_theme_preprocess_page_title(&$variables) {
     $variables['title'] = '';
   }
 }
+
+/**
+ * Implements hook_theme_suggestions_HHOK_alter().
```
amitaibu, line 153:
```suggestion
 * Implements hook_theme_suggestions_page_title_alter().
```

web/themes/custom/server_theme/templates/server-theme-page-title.html.twig
```diff
@@ -0,0 +1,25 @@
...
+ *   displayed in front of the main title tag that appears in the template.
+ * - title: The page title, for use in the actual content.
+ * - title_suffix: Additional output populated by modules, intended to be
+ *   displayed after the main title tag that appears in the template.
+ */
+#}
+{%
+  set classes = [
+    'page-title',
+    'text-5xl',
```
amitaibu, line 18: Maybe 

```suggestion
    'text-lg',
```

amitaibu:
Lovely, thanks.

PR #116, opened 2021-01-12
ddev pull should get qa env

amitaibu:
next I'll add an entry in README


.ddev/import.yaml
```diff
@@ -0,0 +1,3 @@
+provider: pantheon
```
amitaibu, line 1: I'll remove that in favor of entry in README about `ddev config pantheon --pantheon-environment=live`

PR #117, opened 2021-01-13
DDEV ❤ Pantheon

AronNovak:


amitaibu:
Duplicate of https://github.com/Gizra/drupal-starter/pull/116

But anyway, instead we'll explain in the readme devs can use `ddev config pantheon --pantheon-environment=live`

PR #121, opened 2021-01-14
Different containers in styleguide

bboro:
Added output of full, wide and narrow containers and elements on styleguide.

Added fluid-container-full and separated containers into
container.pcss.

Moved footer to full-width container on styleguide.
Changed within the footer the 'wider' container to 'wide'.

Moved user images to narrow width container on styleguide.

image

TB: 0.75h


amitaibu (approved):
Thanks

PR #128, opened 2021-01-29
Add support for PostgreSQL

mariano-dagostino:


amitaibu:
Currently all of our projects are hosted on Pantheon, so there isn't a use for Postgres (which I btw rally like).

What's the use case for this?

mariano-dagostino:
@amitaibu I've been using drupal for more than 10 years and this is the first time I found setup Drupal in a PostgreSQL so easy. In case a someone want to create a patch that needs some special support for PostgreSQL in order to be committed to core (or some contrib) it is nice to have a quick way to setup an instance like this.

amitaibu:
I understand. However I believe it's outside of the scope of this project. IMO it should live in ddev-contrib which is pretty awesome. 

I'd move this info to their explanation about Postgres - https://github.com/drud/ddev-contrib/tree/master/docker-compose-services/postgres

Sorry :smiley_cat:

PR #129, opened 2021-01-30
Elasticsearch updates

AronNovak:
From the latest client project

We should not add stopwords by default, it was a troublemaker for full text search, at least with out of the box views fulltext search.


AronNovak:
@amitaibu This is the first batch of changes. I plan another PR where I add more safety for user provisioning. With the current code, it's too easy to cause a downtime by creating new password for existing users.

amitaibu:
Thanks

amitaibu:
> With the current code, it's too easy to cause a downtime by creating new password for existing users.

btw, how is the PR changing this?

AronNovak:
@amitaibu It will be covered by the next PR.

Issue #132, opened 2021-02-03
Docker rate limit [3h]


amitaibu:
https://blog.travis-ci.com/docker-rate-limits

Seems we're already hitting this from time to time. Let's update travis to help us with this

AronNovak:
https://github.com/travis-ci/travis-ci/issues/5358#issuecomment-542672812
That's a very exciting option.

AronNovak:
As it might reduce the build times too.

AronNovak:
@amitaibu https://hub.docker.com/u/gizra - we'd need to upgrade this (or another Dockerhub account) to Pro ($5/month) and get an access token.

bboro:
Looks like this is done. Closing.

PR #133, opened 2021-02-04
WIP: Docker image cache

AronNovak:
#132


AronNovak:
It's a no-go, it slows down the other stages too, the cache is common:
image

PR #135, opened 2021-02-04
Docker auth

AronNovak:
#132


amitaibu:
Is this working?
How will we deal with new projects - they will need to re-create the hash?

AronNovak:
@amitaibu This did not work yet, my key is not pro, see https://github.com/Gizra/drupal-starter/issues/132#issuecomment-773151697

We need one key, but indeed, `travis encrypt --pro` needs to happen separately for every single project where we want stable builds.

AronNovak:
image

Not yet working, still get this weird retrying process.

AronNovak:
Or that's actually another issue, as rate limiting would look like this: https://github.com/drud/ddev/issues/2628#issue-739843582 :confused:

AronNovak:
@amitaibu (Sadly) we do not need the Docker pro account. It did not solve the problem, but changing to the Google-powered mirror helps.

.travis.yml
```diff
@@ -2,56 +2,56 @@ sudo: required
 language: php
 dist: bionic
 php:
-  - 7.4
-
+- 7.4
```
amitaibu, line 5: let's keep indentation as before (here and below)

.travis.yml
```diff
@@ -2,56 +2,56 @@ sudo: required
...
   global:
-    - GIT_HOST="{{ PANTHEON_GIT_HOST }}"
-    - PANTHEON_GIT_URL="{{ PANTHEON_GIT_URL }}"
-    -
+  - GIT_HOST="{{ PANTHEON_GIT_HOST }}"
+  - PANTHEON_GIT_URL="{{ PANTHEON_GIT_URL }}"
 before_install:
-  - export PATH="$HOME/.config/composer/vendor/bin:$PATH"
-
+- export PATH="$HOME/.config/composer/vendor/bin:$PATH"
```
amitaibu, line 17: Worth a comment on this section, and on the mirror

amitaibu (commented):
Thanks.

1) I don't see a secure access token - shouldn't there be one?
2) So what would be the workflow for a new/ existing project to add that access token?

AronNovak:
For now, we would not need to add the token.
For new, existing projects, all they need to do: https://github.com/Gizra/jep-portal/pull/1219

AronNovak:
For new / existing projects, they just need to add the few new lines in the `before_install` section. Likely a patch from this PR would apply.

AronNovak:
Superseded by our Docker mirror.

PR #140, opened 2021-02-22
Deployment exclusions

AronNovak:


AronNovak:
Seen a strange deployment changeset...

AronNovak:
ping

amitaibu:
I think this had to do with a TW v1 installation we had. I think we should not merge this, since we're no longer supporting that version. Is it ok if I close, or do you think I'm missing your point?

AronNovak:
It's okay, that client project did use TW v1 almost certainly.

PR #141, opened 2021-03-02
Travis optimization

AronNovak:
This way we only spin up DDEV once when the code is merged to master and we want to test and deploy.
Also it seems after we merged something, it's useless to lint.

It"s already used on client projects with satisfaction.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -32,34 +32,41 @@ jobs:
   include:
     - stage: Lint
       name: "PhpStan"
+      if: branch != "master" AND tag IS blank
       script:
         - "composer install"
         - "vendor/bin/phpstan analyse -c phpstan.neon"
     - stage: Lint
       name: "Drupal coding standard: phpcs"
+      if: branch != "master" AND tag IS blank
       script:
         - "$TRAVIS_BUILD_DIR/ci-scripts/install_coder.sh || travis_terminate 1;"
         - "export REVIEW_STANDARD=\"Drupal\" && $TRAVIS_BUILD_DIR/ci-scripts/test_coder.sh"
     - stage: Lint
+      if: branch != "master" AND tag IS blank
       name: "Drupal coding best practices: phpcs"
       script:
         - "$TRAVIS_BUILD_DIR/ci-scripts/install_coder.sh || travis_terminate 1;"
         - "export REVIEW_STANDARD=\"DrupalPractice\" && $TRAVIS_BUILD_DIR/ci-scripts/test_coder.sh"
     - stage: Lint
+      if: branch != "master" AND tag IS blank
       name: "Shell coding standard: shellcheck"
       script:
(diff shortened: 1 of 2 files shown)


.travis.yml
```diff
@@ -32,34 +32,41 @@ jobs:
   include:
     - stage: Lint
       name: "PhpStan"
+      if: branch != "master" AND tag IS blank
```
amitaibu, line 35: I'm not sure about the no-master test part. Sometimes there's a hot fix you might commit directly to master - but we'd like to know if it failed or not.

RoboFile.php
```diff
@@ -589,7 +589,7 @@ public function deployConfigAutodeploy(string $token, string $project_name, $git
       ->to($pantheon_git_host)
       ->run();
     $this->taskReplaceInFile('.travis.yml')
-      ->from('{{ PANTHEON_DEPLOY_BRANCH }}')
+      ->from('master')
```
amitaibu, line 592: Why is this change needed?

.travis.yml
```diff
@@ -32,34 +32,41 @@ jobs:
...
       script:
         - "$TRAVIS_BUILD_DIR/ci-scripts/install_ddev.sh || travis_terminate 1;"
         - "$TRAVIS_BUILD_DIR/ci-scripts/install_drupal.sh || travis_terminate 1;"
         - "$TRAVIS_BUILD_DIR/ci-scripts/test_phpunit.sh || travis_terminate 1;"
     - stage: Deploy
-      name: "Deploy to Pantheon"
-      if: branch = "{{ GITHUB_DEPLOY_BRANCH }}" AND type = push
+      name: "Backend tests: Functional tests and deploy to Pantheon"
+      if: branch = "master" AND type = push
```
amitaibu, line 66: To confirm, change here is that before deploy we do another test? Sounds like a good idea

.travis.yml
AronNovak: We do test on master, but we would not lint.

RoboFile.php
AronNovak: I dropped `{{ PANTHEON_DEPLOY_BRANCH }}` to be coherent. I wanted to have an immediately working `.travis.yml`, if someone doesn't activate auto-deploy, the YML should be usable, that was the idea.
And now we refer this special branch at much more places, at the conditions.

.travis.yml
AronNovak: Yep, well, we do one test. But instead of doing the test and the deploy in two separate jobs, inside one build, now we only launch one job to save time.

.travis.yml
amitaibu: Since it quite quick tests, its worth linting also on `master` - just in case something arrived directly to `master` without passing in a PR first

AronNovak: Ok, I deleted all the conditions for the super lightweight lints.

amitaibu:
Thanks

Issue #138, opened 2021-02-19
Automatic deployment - never go back in time [3h]


AronNovak:
Right now it's possible that automatic deployment reverts already deployed changes.
Deployment jobs might be created like this:
Job A
Job B

But executed in this order:
Job B
Job A

Then changes from Job B are reverted essentially.
To prevent that, we could store the timestamp of the Git commit in the artifact repository to ensure that we never deploy outdated code.

AronNovak:
@amitaibu 
What if we solve this by enabling auto-cancellation of jobs in Travis.
https://blog.travis-ci.com/2017-03-22-introducing-auto-cancellation - we could do it per-branch, for all the branches with auto-deploy, if we activate it, it will never happen again that an older state of the source code gets deployed later than a newer one.
This way we do not need to change anything in the deployment code and we'd eliminate this race condition at another place.

//cc @mariano-dagostino

amitaibu:
Sure, let's use that.

Although I suspect you might still have a race condition, if let's say a newer build start after an previous one has already started; and for some reason the new build is faster.

Anyway, for now I think we can try the auto-cancellation

AronNovak:
> Auto Cancellation allows you to only run builds for the latest commits in the queue. This setting can be applied to builds for Branch builds and Pull Request builds separately. Builds will only be canceled if they are waiting to run, allowing for any running jobs to finish. 

You're right, it's not the perfect solution.

AronNovak:
@amitaibu 
Here's another plan:
 - GH repo: Get `git rev-parse HEAD` - get hash of the current deployment
 - Artifact repo: store the hash in a metadata file, like .deployment
 - Abort the process if the hash in .deployment isn't in the GH Git history

If things go well, commit A, B, C gets deployed sequentially, when B is deployed, hash A is in the Git history.
If things are screwed, A, B, C gets deployed in a random order, let's say B, C, A, the last one would be aborted as hash of "C" is not in the history of the working copy of A.

amitaibu:
Sounds good

ordavidil:
@AronNovak 

I suspect that we still face a race condition with the current solution:

1. Job A cloning Pantheon's repo (invoked by https://github.com/Gizra/drupal-starter/blob/master/.travis.yml#L69)
[...]

PR #144, opened 2021-03-13
Automatic deployment - never go back in time

AronNovak:
#138

### DIFF
--- a/RoboFile.php
+++ b/RoboFile.php
@@ -289,6 +289,13 @@ class RoboFile extends Tasks {
       $currently_deployed_version = trim(file_get_contents($deployment_version_path));
     }
 
+    $result = $this
+        ->taskExec('git rev-parse HEAD')
+        ->printOutput(FALSE)
+        ->run();
+
+    $current_version = trim($result->getMessage());
+
     if (!empty($currently_deployed_version)) {
       $result = $this
         ->taskExec('git cat-file -t ' . $currently_deployed_version)
@@ -296,7 +303,9 @@ class RoboFile extends Tasks {
         ->run();
 
       if ($result->getMessage() !== 'commit') {
-        $this->yell('The commit what is deployed is more recent what we have.');
+        $this->yell(strtr('This current commit @current-commit cannot be deployed, since new commits have been created since, so we don\'t want to deploy an older version.', [
+          '@current-commit' => $current_version,
+        ]));
         $this->yell('Aborting the process to avoid going back in time.');
         return;
       }
@@ -362,12 +371,7 @@ class RoboFile extends Tasks {
     // belong to settings.php.
     $this->_exec("cp web/sites/default/settings.pantheon.php $pantheon_directory/web/sites/default/settings.php");
 
-    $result = $this
-        ->taskExec('git rev-parse HEAD')
-        ->printOutput(FALSE)
-        ->run();
-
-    $current_version = trim($result->getMessage());
+    // Flag the current version in the artifact repo.
     file_put_contents($deployment_version_path, $current_version);
 
     // We don't want to change Pantheon's git ignore, as we do want to commit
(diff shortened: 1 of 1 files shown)


AronNovak:
@amitaibu Shall we try it out at a client project before we merge this or the other way around?

amitaibu:
Sure, we can.

RoboFile.php
```diff
@@ -273,11 +273,35 @@ public function deployPantheon($branch_name = 'master', $commit_message = NULL)
...
+    }
+
+    if (!empty($currently_deployed_version)) {
+      $result = $this
+        ->taskExec('git cat-file -t ' . $currently_deployed_version)
+        ->printOutput(FALSE)
+        ->run();
+
+      if ($result->getMessage() !== 'commit') {
+        $this->yell('The commit what is deployed is more recent what we have.');
```
amitaibu, line 299:
```suggestion
        $this->yell('This current commit @current-commit cannot be deployed, since new commits have been created since, so we don't want to deploy an older version.');
```

amitaibu:
Yeah, I believe Or is right, and we'd need to implement/re-use some locking

amitaibu:
This doesn't seem exactly what we'd need, since it's lacking API - but idea is similar -- http://www.lockasaservice.com/

AronNovak:
Right. My idea is to use the Drupal instance that we deploy to - to do this locking. MySQL locks via `sqlq` are not appropriate, they disappear after closing the session.
We might create a Drush command that obtains a lock, sets a deploy mode variable and we invoke it via Terminus at the very beginning.

AronNovak:
@amitaibu https://github.com/Gizra/drupal-starter/pull/144/commits/f81bb5d0dee82fd984681a69d89e6421d7f9572b - without ironing out the details, that's my plan for now

amitaibu:
I was thinking about a more general purpose lock as a service. What I had in mind was a Google cloud function connected to Cloud SQL - https://cloud.google.com/sql/docs/postgres/connect-functions

Where API could be:

- Try to obtain lock (With optional break time)
- Release lock

AronNovak:
@amitaibu Got the point. But would we want to add one more external dependency that is required in order to deploy?
So far we have Travis and GitHub - and obviously Pantheon.

amitaibu:
I think it's fine, I've deployed a Serverless function 3 years ago and its working with no issue - so I think it's a safe since its serverless, and we don't have to manage it.

amitaibu:
I've done some rethinking on this issue over the weekend, and I think that for now we can probably go with the initial simpler idea - without the locking.

The reason is that currently we auto-deploy to QA, not to LIVE or anything. So we'd save the time for developing the Serverless lock, and accept the (low) risk that we'd still go back on time on QA env.

AronNovak:
@amitaibu so good to merge then as an acceptable trade-off?

amitaibu:
Thanks

RoboFile.php
```diff
@@ -272,127 +272,171 @@ public function deployPantheon($branch_name = 'master', $commit_message = NULL)
       throw new Exception('You need to fill the "PANTHEON_NAME" const in the Robo file. so it will know what is the name of your site.');
     }
 
-    $pantheon_directory = '.pantheon';
+    try {
+      $pantheon_env = $branch_name == 'master' ? 'dev' : $branch_name;
+      $deploy_mode_obtained = $this->_exec("terminus remote:drush " . self::PANTHEON_NAME . "." . $pantheon_env . " server_general:obtain-deploy-mode")->getExitCode();
```
amitaibu, line 277: @AronNovak I think there was a file missing from the PR with the drush command

`server_general:obtain-deploy-mode`

AronNovak: It was an idea that we discarded in https://github.com/Gizra/drupal-starter/pull/144#issuecomment-803555997 , but I forgot to revert the commit where I started to lay off the foundations for it.

PR #147, opened 2021-03-26
Home assignment - BE task

semogilevsky:
1. Installed og, og_ui modules
2. Article CT configured as OG group
3. In the server_theme_preprocess_node hook added subscribe link, condition: current user is not anonymous AND viewed node is group AND current user is not memebr of the group
4. If subscribe link is not empty it is added to node.twig.html template
im001
5. After clicking on the link user is being redirected to subscribe page
img002
6. Added testcase for subscribe link
img003


amitaibu:
Please create this in a new repo, not as PR

PR #150, opened 2021-04-07
Upgrade to ddev 1.17

amitaibu:
cc @AronNovak

### DIFF
--- a/.ddev/commands/web/phpcs
+++ b/.ddev/commands/web/phpcs
@@ -4,4 +4,4 @@
 ## Usage: phpcs
 ## Example: "ddev phpcs"
 
-cd .. && ./vendor/bin/robo phpcs
+./vendor/bin/robo phpcs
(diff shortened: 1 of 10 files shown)


AronNovak:
@amitaibu Pushed a missing bit, looks good otherwise!

AronNovak (approved):

amitaibu:
@AronNovak I'm looking at https://github.com/drud/ddev/blob/master/cmd/ddev/cmd/dotddev_assets/providers/pantheon.yaml.example and I think we need to approach the pantheon auth differently. I'll try it out

amitaibu:
Yeah, I think we'd like to restructure a bit our Robofile, for example instead of defining the Pantheon site name directly in it, we'd extract that info from the [yaml](https://github.com/drud/ddev/blob/e702a98080b459e3fd7742e27dd3400ad46fe5bc/cmd/ddev/cmd/dotddev_assets/providers/pantheon.yaml.example#L40) file

amitaibu:
@AronNovak Marking as needs review, since I think #151 can be added as a follow up

AronNovak (approved):
Just corrected the syntax for `auth:login`, all good otherwise.

amitaibu:
Thanks

PR #157, opened 2021-05-04
Suggestion: Add link to style guide in drupal toolbar.

bserem:
Mobile:
image

Desktop:
image

A follow-up to add permissions to the style-guide might be a good idea, so that the link is only visible if a user can access the style guide.


bboro:
Looks like a nice UX improvement for QA folks. Looks good to me, however I've requested review from Amitai for final say :)

amitaibu:
Thanks

PR #167, opened 2021-06-19
Switch to TW's JIT (Just-In-Time) mode

amitaibu:
https://tailwindcss.com/docs/just-in-time-mode

[Tailwind CSS v2.2](https://blog.tailwindcss.com/tailwindcss-2-2)

### DIFF
--- a/.ddev/commands/web/tw
+++ b/.ddev/commands/web/tw
@@ -1,0 +1,7 @@
+#!/bin/bash
+
+## Description: Run Tailwind CSS compile, watch for changes.
+## Usage: tw
+## Example: "ddev tw"
+
+cd ./web/themes/custom/server_theme && npx postcss ./src/scss/style.pcss --output=./dist/css/style.css --watch
--- a/.ddev/commands/web/tw
+++ b/.ddev/commands/web/tw
@@ -4,4 +4,4 @@
 ## Usage: tw
 ## Example: "ddev tw"
 
-cd ./web/themes/custom/server_theme && npx tailwindcss -i ./src/scss/style.pcss -o ./dist/css/style.css --watch
+cd ./web/themes/custom/server_theme && npx postcss ./src/scss/style.pcss --output=./dist/css/style.css
--- a/.ddev/commands/web/tw
+++ b/.ddev/commands/web/tw
@@ -4,4 +4,4 @@
 ## Usage: tw
 ## Example: "ddev tw"
 
-cd ./web/themes/custom/server_theme && npx postcss ./src/scss/style.pcss --output=./dist/css/style.css --watch
+cd ./web/themes/custom/server_theme && npx tailwindcss -i ./src/scss/style.pcss -o ./dist/css/style.css --watch
--- a/.ddev/commands/web/tw
+++ b/.ddev/commands/web/tw
@@ -4,4 +4,4 @@
 ## Usage: tw
 ## Example: "ddev tw"
 
-cd ./web/themes/custom/server_theme && npx postcss ./src/scss/style.pcss --output=./dist/css/style.css
+cd ./web/themes/custom/server_theme && TAILWIND_MODE=watch npx postcss ./src/scss/style.pcss --output=./dist/css/style.css --watch
--- a/README.md
+++ b/README.md
@@ -50,7 +50,7 @@ On the local development environment, which is using TailWind's [JIT](https://ta
 ddev tw
 ```
 
-This will compile TailWind and keep watching and changes.
+This will compile Tailwind and keep watching for any changes.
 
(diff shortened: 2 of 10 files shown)


RoboFile.php
```diff
@@ -52,10 +52,7 @@ private function compileTheme_($optimize = FALSE) {
     // Make sure we have all the node packages.
     $this->_exec("cd $theme_dir && npm install");
 
-    // If we are asked to optimize, we make sure to purge tailwind's css, by
-    // @link https://tailwindcss.com/docs/optimizing-for-production#removing-unused-css
-    $purge_prefix = $optimize ? 'NODE_ENV=production' : '';
-    $result = $this->_exec("cd $theme_dir && $purge_prefix npx postcss ./src/scss/style.pcss --output=./dist/css/style.css");
+    $result = $this->_exec("cd $theme_dir && npx postcss ./src/scss/style.pcss --output=./dist/css/style.css");
```
amitaibu, line 55: jit mode purges by default

web/themes/custom/server_theme/postcss.config.js
```diff
@@ -6,5 +6,6 @@ module.exports = {
     require('postcss-nested'),
     require('tailwindcss'),
     require('postcss-preset-env')({ stage: 1 }),
+    require('cssnano')({ preset: 'default' }),
```
amitaibu, line 9: We never wired `cssnano` before, so here goes

web/themes/custom/server_theme/server_theme.libraries.yml
```diff
@@ -9,6 +9,7 @@ global-scripts:
     dist/js/scripts.js: {}
   dependencies:
     - core/jquery
+    - core/drupal
```
amitaibu, line 12: unrelated - prevents an error in `scripts.js` (`Drupal` is undefined)

web/themes/custom/server_theme/tailwind.config.js
```diff
@@ -16,9 +17,6 @@ module.exports = {
       '2': '2',
     }
   },
-  variants: {
```
amitaibu, line 19: Not needed, since we use `jit`

nedSaf:
@amitaibu Because JIT adds only the used classes to the styling file, we need to watch all the files that are under `purge -> content`, it needs to add classes to the `style.css` file when the dev adds them in the `twig` file for example.

amitaibu:
Indeed, that should be the case - you now have `ddev tw` for that

.ddev/commands/web/tw
```diff
@@ -0,0 +1,7 @@
+#!/bin/bash
+
+## Description: Run Tailwind CSS compile, watch for changes.
+## Usage: tw
+## Example: "ddev tw"
+
+cd ./web/themes/custom/server_theme && TAILWIND_MODE=watch npx postcss ./src/scss/style.pcss --output=./dist/css/style.css --watch
```
amitaibu, line 7: @nedSaf fyi

nedSaf: Ohh, locally it didn't watch for changes in the twig files or the module files for changes in classes, maybe something is wrong with my config, I'll check later.

amitaibu: If you are trying on another project make sure that:

1. `cd ./web/themes/custom/server_theme` is the correct path after you `ddev ssh`
2. `purge.content` is defined in your `tailwind.config.js` file

README.md
```diff
@@ -42,16 +42,17 @@ then execute the following, and re-try installation steps.
...
+On the local development environment, which is using TailWind's [JIT](https://tailwindcss.com/docs/just-in-time-mode) (Just-In-Time), execute:
 
-On the local development environment, which is using TailWind, execute:
 ```bash
-ddev robo theme:compile-debug
+ddev tw
 ```
 
-This will compile TailWind with all the classes, copy any fonts, images, etc.
+This will compile TailWind and keep watching and changes.
```
bboro, line 53:
```suggestion
This will compile Tailwind and keep watching for any changes.
```

bboro (approved):
Just a little bit of typo

web/themes/custom/server_theme/tailwind.config.js
bboro: @amitaibu I've been thinking about this JIT implementation since yesterday :smile: And this was bugging me a little bit and I wanted to flag in case it's a real issue: when we use `ddev robo theme:compile` (for production) does `jit` mode still work fine without these `variants` defined?

amitaibu: It does, because we define the `mode: jit` on the config file, so the only change really is that we don't invoke `--watch` on `ddev robo theme:compile`

PR #174, opened 2021-07-12
Improve Style guide page

amitaibu:
- Remove unused themes
- Use a custom wrapper for elements
- Collapse in accordion all elements
- If hash is found in URL (e.g. `/style-guide#multiple-cards-with-title`), it will auto-open the pane
- Update PEVB to latest

image


amitaibu:
Code base is now inline with all our recent way of theming.

bboro (approved):
Minor bits I could spot

PR #175, opened 2021-07-15
Update plugins to composer 2 compatible versions.

bserem:
`drupal-starter` already uses Composer 2 for some time now (https://github.com/Gizra/drupal-starter/commit/cec7f8e5c4a40b312c060a43a15e61ee12754506).

So, the only thing left to do here is to require updated versions of two plugins, which would come by default on any new installation anyway.


amitaibu (approved):

amitaibu:
Thanks

amitaibu:
Oh, seems the PR was wrong, as after ddev composer install I get

image

bserem:
I honestly didn't see this coming, since you already have the new versions.

This https://github.com/Gizra/drupal-starter/pull/176 includes the lockfile.

PR #178, opened 2021-07-21
RoboFile improvements

AronNovak:


RoboFile.php
```diff
@@ -423,6 +422,7 @@ public function deployPantheon($branch_name = 'master', $commit_message = NULL)
     // This "git push" above is as async operation, so prevent invoking
     // for instance drush cim before the new changes are there.
     usleep(self::DEPLOYMENT_WAIT_TIME);
+    $pantheon_env = $branch_name == 'master' ? 'dev' : $branch_name;
```
AronNovak, line 425: This was the undefined part.

RoboFile.php
```diff
@@ -603,33 +605,33 @@ public function deployConfigAutodeploy(string $token, string $project_name, $git
         ->getExitCode();
 
       if ($result !== 0) {
-        throw new \Exception('The installation of the dependencies failed.');
```
amitaibu, line 606: Is this change correct?

amitaibu (approved):

RoboFile.php
AronNovak: Yep, we're not inside a namespace, it works well.
An example:
image

PR #180, opened 2021-07-22
Do not do ssh-keyscan for auto-deploy

AronNovak:
On a client project, we faced with this:

```
The authenticity of host '[codeserver.dev.17e5f0a7-6478-4265-98c1-c36a7cdf7083.drush.in]:2222 ([35.224.10.91]:2222)' can't be established.

RSA key fingerprint is SHA256:yPEkh1Amd9WFBSP5syXD5rhUByTjaKBxQnlb5CahZZE.

Are you sure you want to continue connecting (yes/no)? 
```

So it seems to be smarter just to allow whatever we get.
`ssh-keyscan` anyways do not protect us from MITM attacks.


amitaibu:
Thanks

Issue #184, opened 2021-07-29
Create ddev commands for theme compiling and watching


andreytroeglazov:
## Issue
In current setup we have 2 commands for theme - ddev tw(theme watching) and ddev robo theme:compile(theme compiling), I suggest to make 2 ddev commands which will have similar namings - ddev theme:watch and ddev theme:compile

bserem:
@andreytroeglazov  Can we close this?

PR #185, opened 2021-08-02
New theme commands

andreytroeglazov:
#184
- Changed watching command to theme:watch - ddev theme:watch
- Added new command for compiling theme:compile - ddev theme:compile
- Updated readme

### DIFF
--- a/.ddev/commands/web/theme:watch
+++ b/.ddev/commands/web/theme:watch
@@ -1,7 +1,7 @@
 #!/bin/bash
 
 ## Description: Run Tailwind CSS compile, watch for changes.
-## Usage: tw
-## Example: "ddev tw"
+## Usage: theme:watch
+## Example: "ddev theme:watch"
 
 cd ./web/themes/custom/server_theme && TAILWIND_MODE=watch npx postcss ./src/pcss/style.pcss --output=./dist/css/style.css --watch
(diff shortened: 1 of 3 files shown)


andreytroeglazov:
Tested on jep-portal
Selection_517
Selection_516

.ddev/commands/web/theme:watch
```diff
@@ -1,7 +1,7 @@
 #!/bin/bash
 
 ## Description: Run Tailwind CSS compile, watch for changes.
-## Usage: tw
-## Example: "ddev tw"
+## Usage: theme:watch
```
amitaibu, line 4: As we will be watching only Tailwind, and not new images/fonts/JS. Maybe we should call `theme:watch:tw`? Or keep simple and leave as is?

andreytroeglazov: `theme:watch:tw` imo its a little bit complex and also we wont have `theme:watch`
I think its worth to update readme to let developers know that its watching only styles and leave it as `theme:watch`.

amitaibu: > I think its worth to update readme to let

Sounds good, go for it :)

andreytroeglazov: @amitaibu readme is updated now.

amitaibu:
Thanks

PR #192, opened 2021-08-31
Enable CSS & JS aggregation by default for non-DDEV environments

AronNovak:
Right now for every single client project, one day we need to change the config YAML.


AronNovak:
image
Checked, but overridden

amitaibu (approved):

PR #195, opened 2021-09-03
Full HTML restrictions

AronNovak:


AronNovak:
image

Still we are not restricted too much, only `script` and `iframe` should be disabled entirely (on top of `on..` JS event attributes)

AronNovak:
@amitaibu What do you think? Is it overly specific or ugly to be included? Do you generally agree that we should do something with Full HTML to exclude known dangerous stuff?

amitaibu:
@AronNovak what did you use to get the list of allowed HTML values?

AronNovak:
AFAIK https://www.w3schools.com/tags/default.asp and https://www.w3schools.com/tags/ref_attributes.asp

PR #198, opened 2021-09-13
Use PostRequestIndexingTrait instead of search_api_cron()

bboro:
Also add a search test base class with waitForElasticSearchIndex() method.


web/modules/custom/server_general/tests/src/ExistingSite/ServerGeneralSearchTestBase.php
```diff
@@ -0,0 +1,53 @@
+<?php
+
+declare(strict_types = 1);
```
AronNovak, line 3: https://www.php.net/manual/en/language.types.declarations.php#language.types.declarations.strict
:100: 
Just learned something new!

AronNovak (approved):

PR #199, opened 2021-09-24
Do not set hard-wired admin pw by default

AronNovak:
As at the beginning, we might push QA database to `live`, this seems to be a bad idea - to do it by default for every site.


amitaibu (approved):
Thanks. Travis is red, though.

AronNovak:
@amitaibu Yep, but it's a generic Travis issue, working on the fix (or at least a temporary workaround) first for an in-progress client project.
It seems simply git.drupacode.org doesn't behave nicely.

PR #201, opened 2021-09-30
Don't deploy dev dependencies to prod, and tidy up the RoboFile

bboro:
Main change:

```php
    // Note: Add web/libraries to cleanup list below if the project installs it.
    // This drupal starter currently doesn't install any.
    $this->taskExec("rm -rf vendor web/core web/modules/contrib && composer install --no-dev")->run();
```

And moved `totten/lurkerlite` package from dev dependency to non-dev.

The rest is just cleanup of phpdocs, and adding return types to functions and small refactors.

Time spent: 0.8h

### DIFF
--- a/RoboFile.php
+++ b/RoboFile.php
@@ -1,6 +1,9 @@
 <?php
 
+declare(strict_types = 1);
+
 use Lurker\Event\FilesystemEvent;
+use Robo\ResultData;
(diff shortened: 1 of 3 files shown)


AronNovak (approved):

amitaibu:
Thanks

Issue #209, opened 2021-10-29
Add stage_file_proxy module


andreytroeglazov:
It will be useful for local development in case a developer does not want to download all files.

andreytroeglazov:
Done.

PR #210, opened 2021-10-29
Stage file proxy

andreytroeglazov:
#209

### DIFF
--- a/README.md
+++ b/README.md
@@ -245,7 +245,3 @@ Optionally you can specify which target branch you'd like to push on Pantheon, b
 If you don't want to copy production files locally, you can enable stage_file_proxy module.
 It saves you time and disk space by sending requests to your development environment's files directory
 to the production environment and making a copy of the production file in your development site.
-
-This module shouldn't be enabled on live, to exclude it from configuration, put code below to your settings file.
-
-    $settings['config_exclude_modules'] = [ 'stage_file_proxy'];
--- a/README.md
+++ b/README.md
@@ -245,3 +245,5 @@ Optionally you can specify which target branch you'd like to push on Pantheon, b
 If you don't want to copy production files locally, you can enable stage_file_proxy module.
 It saves you time and disk space by sending requests to your development environment's files directory
 to the production environment and making a copy of the production file in your development site.
+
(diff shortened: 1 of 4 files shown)


README.md
```diff
@@ -240,3 +240,12 @@ Optionally you can specify which target branch you'd like to push on Pantheon, b
 
     # Pull DB & Files
     ddev pull pantheon
+
+## Stage File Proxy
+If you don't want to copy production files locally, you can enable stage_file_proxy module.
+It saves you time and disk space by sending requests to your development environment's files directory
+to the production environment and making a copy of the production file in your development site.
+
+This module shouldn't be enabled on live, to exclude it from configuration, put code below to your settings file.
```
amitaibu, line 249: We manage [Pantheon](https://github.com/Gizra/drupal-starter/blob/main/web/sites/default/settings.pantheon.php) and [DDEV](https://github.com/Gizra/drupal-starter/blob/main/web/sites/default/settings.ddev.php) `settings.php` files in this repo, so let's add it there

andreytroeglazov: @amitaibu changed readme and added exclusion to settings.ddev.php

README.md
```diff
@@ -240,3 +240,8 @@ Optionally you can specify which target branch you'd like to push on Pantheon, b
 
     # Pull DB & Files
     ddev pull pantheon
+
+## Stage File Proxy
+If you don't want to copy production files locally, you can enable stage_file_proxy module.
```
amitaibu, line 245: Let's enable it by default on ddev. Maybe by enabling it in the local config file - https://github.com/Gizra/drupal-starter/blob/cd7aa5c47bd3b9ea420d025b77c5b7680411fa35/.ddev/config.local.yaml.example#L23

andreytroeglazov: @amitaibu not sure it's needed, because you need to configure production path to make it work, just enabling module in ddev won't give any effect, so if developer wants to use it he can enable & configure it. What do you think?

amitaibu: Ok, so let's add a comment about the need to configure it as-well (e.g. `configure the path at /admin/config/...`)

andreytroeglazov: @amitaibu done

amitaibu (approved):
Thanks

PR #212, opened 2021-11-08
GitPod

AronNovak:
#211 

https://gitpod.io/

<a href="https://gitpod.io/#https://github.com/Gizra/drupal-starter/pull/212"></a>


AronNovak:
image
Obfuscated on purpose? Maybe, but this is a bit odd.

AronNovak:
image

AronNovak:
image
Somehow DDEV is not healthy inside GitPod yet.

AronNovak:
image
Almost here....

amitaibu:
Looks great. A couple of things:

1. Are we able to open the new site in a new browser in the end of the installation?
2. Current when doing `ddev drush uli` I still see the local URL

image

AronNovak:
> Are we able to open the new site in a new browser in the end of the installation?

Based on the doc, we only able to open the preview:
```
  3     command: |
  4       .ddev/gitpod-setup-ddev.sh
  5       ddev composer install
  6       gp await-port 8888 && gp preview $(gp url 8888)
```

What should already happen.

AronNovak:
image
So that's how it looks like with the auto-opened preview.

amitaibu:
One more thing if you can please confirm - is Mailhog working on Gitpod?

AronNovak:
@amitaibu 
image

Not yet, let me see if I can force HTTP.

AronNovak:
image
It does, we can simply pick the non-HTTPS version.

amitaibu:
Wonderful, can you please add an entry about it in the README (at the beginning of the doc) + a Gitpod badge for `master`

.ddev/gitpod-setup-ddev.sh
```diff
@@ -0,0 +1,40 @@
...
+additional_fqdns:
+- 8888-${shortgpurl}
+- 8025-${shortgpurl}
+- 8036-${shortgpurl}
+CONFIGEND
+
+# Forces proper external base URL.
+DRUPAL_BASE=$(gp url 8888)
+cat <<DRUSH_CMD > ~/.ddev/commands/web/drush
+  ./vendor/bin/drush --uri="$DRUPAL_BASE" $@
```
AronNovak, line 30: @amitaibu https://github.com/drush-ops/drush/issues/3121 and other issues, I wasn't able to alter the uri via a YAML config file :disappointed:

amitaibu: Ok. Can we have the last command be something like

 ```
drush ddev uli --uri `gp url 8880`
```

Comment (AronNovak) [reply .ddev/gitpod-setup-ddev.sh:30]: We can, but then, all the other commands will contain a misleading URL, not? At least it won't be coherent.

Comment (amitaibu) [reply .ddev/gitpod-setup-ddev.sh:30]: Maybe we can add it inside a [drush.yaml](https://www.drush.org/latest/examples/example.drush.yml/) ?

There was also the php version of it, something that made `$aliases` work

Comment (AronNovak) [reply .ddev/gitpod-setup-ddev.sh:30]: I tried `drush.yaml` at various places, nothing worked so far, I'll give it one more attempt.

Comment (AronNovak) [reply .ddev/gitpod-setup-ddev.sh]: Well, it does work, what's very misleading that it's ignored under `sites/default`

**Comment by AronNovak** [comment]
image

**Comment by AronNovak** [comment]
^^ @amitaibu Here it is, thanks for insisting on the proper fix for the Drush alias.

**Comment by AronNovak** [comment]
Also it's contributed back: https://github.com/shaal/ddev-gitpod/pull/93/files

File: .gitpod.Dockerfile [hunk 2]
```diff
@@ -0,0 +1,14 @@
...
+SHELL ["/bin/bash", "-c"]
+
+RUN sudo apt-get -qq update
+# Install required libraries for Projector + PhpStorm
+RUN sudo apt-get -qq install -y python3 python3-pip libxext6 libxrender1 libxtst6 libfreetype6 libxi6 telnet netcat
+# Install Projector
+RUN pip3 install projector-installer
+# Install PhpStorm
+RUN mkdir -p ~/.projector/configs  # Prevents projector install from asking for the license acceptance
+RUN projector install 'PhpStorm 2021.1.4' --no-auto-run
```
amitaibu, line 11: What is this doing?

amitaibu (approved):

.gitpod.Dockerfile
AronNovak: https://github.com/shaal/ddev-gitpod/blob/main/.gitpod.Dockerfile#L11 - just copied.
I guess you can use PhpStorm inside GitPod, but I did not try it out. We might speed up the init by removing this, i guess.

amitaibu: Let's keep for now, might be handy

PR #222, opened 2022-01-03
Form - theme suggestions

AronNovak:
I guess it makes sense to make it possible to override theme for forms by default, per form.

### DIFF
--- a/web/modules/custom/server_general/server_general.module
+++ b/web/modules/custom/server_general/server_general.module
@@ -10,8 +10,8 @@
  *
  * Adds fine-grained overrides for form templates.
  *
- * For instance for a form with an ID "server_general_moo_form", you can use
- * form--server-general-moo-form.html.twig
+ * For instance for a form with an ID "server_general_foo_form", you can use
+ * form--server-general-foo-form.html.twig
  *
  * That can be based on form.html.twig.
  */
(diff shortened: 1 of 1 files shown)


web/modules/custom/server_general/server_general.module
```diff
@@ -4,3 +4,15 @@
...
  * Module file.
  */
+
+/**
+ * Implements hook_theme_suggestions_alter().
+ */
+function server_general_theme_suggestions_alter(array &$suggestions, array $variables) {
+  if (isset($variables['element']) && isset($variables['element']['#type']) && $variables['element']['#type'] == 'form') {
+    $original_theme_hook = $variables['theme_hook_original'];
+
```
AronNovak, line 14:
```suggestion
```

web/modules/custom/server_general/server_general.module
```diff
@@ -4,3 +4,14 @@
  * @file
  * Module file.
  */
+
+/**
+ * Implements hook_theme_suggestions_alter().
```
amitaibu, line 9: Worth a comment with an example of what the twig name would be.

web/modules/custom/server_general/server_general.module
```diff
@@ -4,3 +4,21 @@
...
+ *
+ * Adds fine-grained overrides for form templates.
+ *
+ * For instance for a form with an ID "server_general_moo_form", you can use
+ * form--server-general-moo-form.html.twig
```
amitaibu, line 14:
```suggestion
 * For instance for a form with an ID "server_general_foo_form", you can use
 * form--server-general-foo-form.html.twig
```

amitaibu (approved):

PR #233, opened 2022-02-07
Install coder 8.3.14 following latest config instructions

mariano-dagostino:
According to this: https://www.drupal.org/node/3260071 it is no longer necessary to specify `installed_paths` for coder to work.


amitaibu:
Thanks

PR #231, opened 2022-02-07
Add Prettier support to the theme folder

amitaibu:
https://tailwindcss.com/blog/automatic-class-sorting-with-prettier

- Added `ddev theme:prettier` command to execute prettier.

@AronNovak when you have time, could please add a linter on Travis that would `cd ./web/themes/custom/server_theme && npx prettier --check .`

PR #236, opened 2022-02-08
Revert "Add Prettier support to the theme folder"

amitaibu:
Reverts Gizra/drupal-starter#231

@AronNovak Now that twig integration is removed, I think it would make more sense to remove the entire feature


AronNovak:
Right, twig integration is not mature enough, it would become too annoying.

AronNovak (approved):

PR #239, opened 2022-02-21
Set theme to 1440px max width and set 80px gutter on XL breakpoint

bboro:


web/themes/custom/server_theme/tailwind.config.js
```diff
@@ -3,6 +3,9 @@ const plugin = require('tailwindcss/plugin');
 module.exports = {
   theme: {
     extend: {
+      maxWidth: {
```
amitaibu, line 6: I think we should define it under `screens` like they do [here](https://tailwindcss.com/docs/theme)

bboro: No that's changing breakpoints right? We don't wanna do that really. This is just changing the max width of the wide container, and applying 80px gutter at XL breakpoint (1280px) and up

amitaibu (approved):
:+1: Thanks

PR #240, opened 2022-02-24
Improve the styleguide UX

bboro:


bboro:
- Stop clicks on styleguide titles jumping the page (event.preventDefault()).
- Initialize slick sliders when they're expanded
- Allow elements to be opened and closed instead of having only 1 element open. This combined with point 1 also stops the clicks on titles triggering the page scrolling, which can trigger some header behaviors etc.

amitaibu (approved):
Thanks :+1:

bboro:
@amitaibu don't merge yet, noticed we don't have slick on drupal-starter by default.

bboro:
- Disabled advagg on ddev
- Restored hiding all panels on styleguide on load
- Added detection of slick library before attempting to call it.

bboro:
- Update the URL hash when titles are clicked (this was disabled by `event.preventDefault()`) - @andreytroeglazov thanks for highlighting this and sorry for not understanding what you meant :D

PR #245, opened 2022-03-25
Add some basic styles to theme user related pages

mariano-dagostino:
# Problem

By using tailwind, there are not default styles for login passwords. This is usually forgotten in projects and creates a really bad UX for users that cannot find the login button.

Screenshot_2022-03-25_11-55-56

# Proposed solution

Let's add a minimal basic styles to prevent this

Screenshot_2022-03-25_11-57-16
Screenshot_2022-03-25_11-57-03

### DIFF
--- a/web/themes/custom/server_theme/src/pcss/style.pcss
+++ b/web/themes/custom/server_theme/src/pcss/style.pcss
@@ -2,9 +2,9 @@
 @import "tailwindcss/components";
 
 @import "container.pcss";
-@import "typography.pcss";
-@import "list.pcss";
 @import "drupal.pcss";
+@import "list.pcss";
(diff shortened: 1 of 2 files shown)


web/themes/custom/server_theme/src/pcss/style.pcss
```diff
@@ -4,6 +4,7 @@
 @import "container.pcss";
 @import "typography.pcss";
 @import "list.pcss";
+@import "drupal.pcss";
```
amitaibu, line 7: Let's order by alphabet

amitaibu (approved):
Thanks, minor comment.

amitaibu:
Thanks :+1:

PR #250, opened 2022-04-11
Add Devel and enable webprofiler

amitaibu:
image

### DIFF
--- a/web/sites/default/settings.ddev.php
+++ b/web/sites/default/settings.ddev.php
@@ -75,3 +75,5 @@ $config['advagg.settings']['enabled'] = FALSE;
 // Excludes stage_file_proxy module from configuration export,
 // as it should not be enabled on production.
 $settings['config_exclude_modules'] = ['stage_file_proxy'];
+
+$settings['config_exclude_modules'] = ['devel', 'webprofiler', 'stage_file_proxy'];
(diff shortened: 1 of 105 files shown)


web/sites/default/settings.ddev.php
```diff
@@ -72,6 +72,10 @@
 // Disable Advagg on ddev.
 $config['advagg.settings']['enabled'] = FALSE;
 
-// Excludes stage_file_proxy module from configuration export,
-// as it should not be enabled on production.
-$settings['config_exclude_modules'] = ['stage_file_proxy'];
+// Excludes modules from configuration export, as they should not be enabled on
+// production.
+$settings['config_exclude_modules'] = [
```
amitaibu, line 77: Exclude from config export

.ddev/config.local.yaml.example
```diff
@@ -28,5 +28,8 @@ hooks:
     - exec: drush sapi-c
     - exec: drush sapi-i
 
+    # Enable Devel and Webprofiler
+    - exec: drush pm-enable webprofiler -y
```
amitaibu, line 32: Enabled here

amitaibu:
@AronNovak not sure why all the config diff - maybe new Drupal version?

.ddev/config.local.yaml.example
AronNovak: :+1:

AronNovak:
@amitaibu Yep, we did several bumps by just `ddev composer`, it seems to be more or less normal.

AronNovak (approved):

PR #254, opened 2022-05-03
Failing patches should not remain silent

AronNovak:
https://github.com/cweagans/composer-patches/blob/1.x/src/Patches.php#L285

### DIFF
--- a/.ddev/config.yaml
+++ b/.ddev/config.yaml
@@ -14,7 +14,8 @@ nfs_mount_enabled: false
 mutagen_enabled: false
 use_dns_when_possible: true
 composer_version: "2"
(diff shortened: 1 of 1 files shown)


amitaibu (approved):
Makes sense :+1: 

Sounds like it could be contributed to DDEV itself.

AronNovak:
Credits to https://github.com/Sweetchuck for the idea on #drupalhu slack :)

AronNovak:
https://github.com/drud/ddev/pull/3795#issuecomment-1116750261 - it seems DDEV won't change the default immediately, but perhaps in the next major release.

PR #269, opened 2022-06-28
Fix deployment rsync ignores

AronNovak:


RoboFile.php
```diff
@@ -364,7 +364,6 @@ public function deployPantheon(string $branch_name = 'master', ?string $commit_m
       '.phpunit.result.cache',
       '.travis.yml',
       'ci-scripts',
-      'drush',
```
amitaibu, line 367: Drush should be deployed?

amitaibu (approved):

RoboFile.php
AronNovak: Yep, as otherwise we cannot use Drush 11, as Pantheon only supports it as site-local version.

bboro: I had a suspicion that this folder contains only ddev's drush related config (drush.yml) which is the reason this was added to the ignore list in the first place. However thinking again, `rsync --exclude=drush` may ignore a lot more things than just the drush folder in project root. Maybe the solution is to change this to `drush/drush.yml` or even `./drush/drush.yml`

RoboFile.php
```diff
@@ -387,6 +386,12 @@ public function deployPantheon(string $branch_name = 'master', ?string $commit_m
       'web/sites/simpletest',
       'web/sites/README.txt',
       'web/themes/README.txt',
+      'web/themes/custom/server_theme/src',
+      'web/themes/custom/server_theme/node_modules',
+      'web/themes/custom/server_theme/package.json',
+      'web/themes/custom/server_theme/package-lock.json',
+      'web/themes/custom/server_theme/tailwind.config.js',
+      'web/themes/custom/server_theme/postcss.config.js',
```
bboro, line 394: :+1: :100:

RoboFile.php
AronNovak: `./drush/drush.yml` does not result in what we want

PR #270, opened 2022-06-28
Ensure `deployPantheon` failures are reported as errors

bboro:
If `git push` fails on the pantheon repo, the `finally {}` part of the code runs `git checkout $original_branch` which exits with code `0`, so the deploy continues running, and runs `deployPantheonSync`.

Since `finally` runs everytime, the only way I could think of was to utilize `exit()` in the `catch` to stop it from running if we get a failure. Although that kinda beats the purpose of `finally` but I haven't got the headspace right now to figure out a better solution :smile: Suggestions welcome.

### DIFF
--- a/RoboFile.php
+++ b/RoboFile.php
@@ -257,6 +257,10 @@ class RoboFile extends Tasks {
     catch (Exception $e) {
       $this->yell('The deployment failed', 22, 'red');
       $this->say($e->getMessage());
+      // Ensure we exit with error as we're catching this exception.
+      $this->taskExec("git checkout $original_branch && exit 1")->run();
+      // Don't run the "finally {}" block as it will exit with code 0.
+      exit();
     }
     finally {
       $this->taskExec("git checkout $original_branch")->run();


AronNovak (commented):
Let's drop the `finally` branch. This happened on dev locals in the past, so restoring the original status made much more sense. Deploying from local is the exception, not the norm, so let's simply get rid of the problematic `finally`, I'd say.

RoboFile.php
```diff
@@ -251,16 +244,21 @@ public function deployTagPantheon(string $tag, string $branch_name = 'master', ?
...
       $this->yell('The deployment failed', 22, 'red');
       $this->say($e->getMessage());
+      // Set exit code to 1 (error).
+      $exit = 1;
     }
-    finally {
-      $this->taskExec("git checkout $original_branch")->run();
-    }
+    // Check out the original branch regardless of success or failure.
+    $this->taskExec("git checkout -")->run();
```
bboro, line 259: We don't really need `$original_branch`. Just `git checkout -` should suffice.

AronNovak (approved):

PR #274, opened 2022-07-26
Add a Media Embed process plugin base class

bboro:


web/modules/custom/server_migrate/src/Plugin/migrate/process/MediaEmbedProcessPluginBase.php
```diff
@@ -0,0 +1,366 @@
...
+        '@error' => $e->getMessage(),
+      ]);
+    }
+
+    // Copy the remote file to local.
+    if (!$file instanceof FileInterface) {
+      try {
+        $plugin_value = NULL;
+        // Try local copy if present.
+        $local_bases = ['public://oistmedia-files/', 'public://'];
```
amitaibu, line 297: Needs adaptation

amitaibu (commented):
Let's add an entry to the README.

amitaibu:
Asking @AronNovak for a code review as well

AronNovak (commented):
Can we somehow utilize this via demo migration? If we leave it hanging in the air, sooner or later it will become broken and we won't notice it.

web/modules/custom/server_migrate/src/Plugin/migrate/process/IframeToMedia.php
```diff
@@ -0,0 +1,174 @@
...
+   *
+   * @param string $url
+   *   The oembed url.
+   *
+   * @return \Drupal\media\MediaInterface
+   *   The media entity.
+   */
+  protected function loadOrCreateVideoMediaByUrl(string $url): MediaInterface {
+    // Remove querystrings and anything that follows them.
+    $url = preg_replace('/\?.*$/', '', trim($url));
```
amitaibu, line 146: Maybe working with `parse_url` would make it easier to read/ debug? Regex always scares me a bit :)

bboro: This is fine, it's not the scariest regex in the class room :)

bboro:
> Can we somehow utilize this via demo migration? If we leave it hanging in the air, sooner or later it will become broken and we won't notice it.

Not to sound defeatist, but this will take too much time and won't be worth it I think. It'll be easier to adapt the migrations per-project and use this as a starting point.

amitaibu:
Thanks

PR #275, opened 2022-07-27
Redis flood flush command

AronNovak:


README.md
```diff
@@ -264,3 +264,29 @@ It saves you time and disk space by sending requests to your development environ
...
+```
+
+It purges all the entries from `test` environment.
+
+```bash
+ddev pantheon-flood-flush test 193.165.2.3
+ddev ddev-flood-flush 193.165.2.3
+```
+
+It purges entries related to `193.165.2.3` from `test` environment, or alternatively from DDEV's own Redis.
```
amitaibu, line 292:
```suggestion
Purges entries related to IP `193.165.2.3` from Pantheon's `test` environment, or alternatively from DDEV's own Redis.
```

README.md
```diff
@@ -264,3 +264,29 @@ It saves you time and disk space by sending requests to your development environ
 to the production environment and making a copy of the production file in your development site.
 
 Configure the origin path at `/admin/config/system/stage_file_proxy`.
+
+### Flood control
+
+As the project uses Redis, it is not possible to use the SQL console to reset flood table.
+There's a custom DDEV command to help with that.
+Usages:
```
amitaibu, line 272:
```suggestion
There's a custom DDEV command to help with that. Usages:
```

web/sites/default/settings.ddev.php
```diff
@@ -66,6 +66,7 @@
 $settings['redis.connection']['host']      = 'redis';  // Your Redis instance hostname.
 $settings['cache']['default'] = 'cache.backend.redis';
 $settings['container_yamls'][] = 'modules/contrib/redis/redis.services.yml';
+$settings['container_yamls'][] = 'modules/contrib/redis/example.services.yml';
```
amitaibu, line 69: Is this one correct?

AronNovak: Yep, that''s the correct one, we should drop the previous line:
```
Either include the default example.services.yml from the module, which will
replace all supported backend services (that currently includes the cache tags
checksum service and the lock backends, check the file for the current list)
or copy the service definitions into a site specific services.yml.

    $settings['container_yamls'][] = 'modules/redis/example.services.yml';
```

AronNovak: Hm, if Redis module is not enabled, we need that live, Travis says so :)

amitaibu: Ok, so we should leave as is? If so, please merge

PR #288, opened 2022-08-29
Add role for UID1 user

AronNovak:
Otherwise all role-based permission check needs to handle the edge case of UID1 that does not have any role but it's supposed to be able to do anything.

Credits to @andreytroeglazov , we concluded that this would be helpful on a client project.

<a href="https://gitpod.io/#https://github.com/Gizra/drupal-starter/pull/288"></a>

### DIFF
--- a/.ddev/config.local.yaml.example
+++ b/.ddev/config.local.yaml.example
@@ -31,5 +31,9 @@ hooks:
     # Enable Devel and Webprofiler
     - exec: drush pm-enable webprofiler -y
 
+    # Add administrator role to admin. Some role-only checks would choke
+    # for super admin otherwise.
+    - exec: drush user-add-role "administrator" admin
+
     # Generate one-time login link
     - exec: drush uli


amitaibu:
IMO this should be `won't fix`, as we should check permissions and not roles. Is there a good use case for that?

AronNovak:
@amitaibu Any client project is full of role-based checks, as I remember, so it's pretty much everywhere. We never paid too much attention to always create a permission and always use that.

One open-sourced client project example:
```
fgrep -R in_array * | grep roles
server/hedley/modules/custom/hedley_admin/hedley_admin.module:  if (in_array('administrator', $account->roles)) {
server/hedley/modules/custom/hedley_admin/hedley_admin.module:  if (!in_array('E-ledger viewer', $account->roles) && !in_array('nurse', $account->roles)) {
```

But other projects are no different.

amitaibu:
I see. Sorry, still think it's wrong. So it's better to have permissions o your projects/ fix existing ones rather than port it to Drupal-starter. Checking permissions already takes care of checking UID 1. Also, checking permissions on OG, could check permissions on the site-wide. In other words, going with role-based checks isn't natural to Drupal.

PR #291, opened 2022-08-31
Restrict creation of user by default and add tests to prevent harmful permissions set to anonymous users

mariano-dagostino:
#290

<a href="https://gitpod.io/#https://github.com/Gizra/drupal-starter/pull/291"></a>


amitaibu (approved):
:+1: Thanks

PR #298, opened 2022-09-07
Remove GitPod integration

amitaibu:
I don't think anyone was using it, so one less thing to maintain


AronNovak (approved):
I agree, for me, it was a bit painful to do all these in the browser honestly.

Issue #259, opened 2022-05-24
Set homepage after migration


amitaibu:
## Issue

Upon clean installation we set the homepage to be `/homepage` on `/admin/config/system/site-information` . But if we check `\Drupal::service('path.matcher')->isFrontPage();` it will return False. If we then manually change the URL and then back to `/homepage` and click save, then under the hood it changes it to the internal path (e.g. /node/21).  

![Selection_999(1065)](https://user-images.githubusercontent.com/125707/170026708-ce8f61b1-1a55-4c07-93fb-561d7c356cc2.jpg)

## Task

1. Confirm my findings
2. Instead of setting it via config, after the migration is executed, we can set it.

bboro:
> Confirm my findings

Yep that's correct. It also keeps forwarding the user to `/homepage` instead of `/` when you go to the root.

> Instead of setting it via config, after the migration is executed, we can set it.

Assume automatically and not manually? I also think we should ignore this configuration so people can freely change this on other environments such as live, qa etc.

PR #304, opened 2022-09-15
Set Homepage config after migration run

bboro:
#259 

- Revert homepage back to default:
   image
- Install `config_ignore` and configure it to **not revert** any changes to the homepage configuration
- Created drush command in `server_general` to set homepage to landing page with name "Homepage":  `drush set-homepage`
- Call the drush command from `config.local.yaml.example` after the migrations run. See below: "Homepage set to /node/7"

image

Homepage output:

image

TB: 1.5h

### DIFF
--- a/config/sync/config_ignore.settings.yml
+++ b/config/sync/config_ignore.settings.yml
@@ -1,0 +1,3 @@
+_core:
+  default_config_hash: UVH1aJ4b44UM-VdPVN7hNNuuVqfReJxwfVeDQH1Hvsk
+ignored_config_entities: {  }
--- a/config/sync/config_ignore.settings.yml
+++ b/config/sync/config_ignore.settings.yml
@@ -1,3 +1,4 @@
 _core:
   default_config_hash: UVH1aJ4b44UM-VdPVN7hNNuuVqfReJxwfVeDQH1Hvsk
-ignored_config_entities: {  }
+ignored_config_entities:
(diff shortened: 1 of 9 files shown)


config/sync/config_ignore.settings.yml
```diff
@@ -0,0 +1,4 @@
+_core:
```
amitaibu, line 1: I think we can remove this part. As in all our projects we keep the `Homepage`. Any other implementor may add this if they want.

bboro: Sorry I don't quite understand what you mean here. Could you clarify?

amitaibu: Sure, I think we remove the `config_ignore` code from the PR - as we always seem to keep the homepage as `/homepage`.

bboro: But this PR specifically fixes that as setting the homepage to `/homepage` doesn't work, it causes forwarding to `/homepage` instead of `/` and when you re-save the config it will set the homepage to `/node/12` for example, the exact ID of the homepage node will change between installs too I think. So thats why we're ignoring this config. Also this allows clients to freely change their homepage to a different landing page without the config getting reverted.

amitaibu: Oh, got it. Thanks.

amitaibu (approved):

PR #311, opened 2022-09-23
Add Intl date module and EntityDateTrait for handling international dates

bboro:


bboro:
No usages at the moment, didn't enable the module. We can enable it as when needed and use the trait when needed.

amitaibu:
> No usages at the moment, didn't enable the module.

Let's wire it to [NodeNews](https://github.com/Gizra/drupal-starter/blob/c1fb89a9d5ba55ced78b7922559ae517b8533656/web/modules/custom/server_general/src/Plugin/EntityViewBuilder/NodeNews.php#L31), and show the New's date.

As we've solved the presentation of dates in a few projects already, I suggest we copy over that code as well. So we cover for example Russian, Chinese, Arabic -- that require special casing.

bboro:
image

TB: +0.7h

AronNovak (commented):
:tada:

PR #312, opened 2022-09-24
Use Tailwind Prose to style text

amitaibu:
TW's [Prose](https://tailwindcss.com/docs/typography-plugin) takes care of styling text that would be under long text fields such as the Body field. We can still [override](https://tailwindcss.com/docs/typography-plugin#element-modifiers) it's classes as needed.

- Upgrade postcss, tailwind and related packages to the latest
- Add a `server_theme_prose_text` to add prose classes, and add a `::wrapElement` for it. My idea is that we could have different Prose for different view modes (e.g. on a card, we'd style it differently)
- Move non-prose h1 - h5 to be under `:not(prose)`

image


amitaibu:
To clarifry the new wrapElement for Prose:

## Before

```php
$element = $this->buildProcessedText($entity);
$build[] = $this->wrapElementWideContainer($element);
```

## After

```php
$element = $this->buildProcessedText($entity);
$element = $this->wrapElementProseText($element);
$build[] = $this->wrapElementWideContainer($element);
```

web/themes/custom/server_theme/templates/server-theme-prose-text.html.twig
```diff
@@ -0,0 +1,3 @@
+<div class="prose md:prose-lg lg:prose-xl prose-headings:text-gray-700">
```
amitaibu, line 1: `prose-headings:text-gray-700` is an example of how we'd override default prose settings

amitaibu:
Closed in favor of https://github.com/Gizra/drupal-starter/pull/319

PR #314, opened 2022-09-24
Add filter empty tags module

amitaibu:
If content editor hits a few enters, they would not see empty paragraphs.

Based on top of https://github.com/Gizra/drupal-starter/pull/312

|Node edit | Before | After |
| -- | -- | -- |
| image | image | image


amitaibu:
@isapisa fyi ^^

Issue #315, opened 2022-09-24
`system.site:page.front` isn't ignored


amitaibu:
In https://github.com/Gizra/drupal-starter/pull/304 `system.site:page.front` was added to config ignore, and it shows on the settings

![image](https://user-images.githubusercontent.com/125707/192114516-0270d230-8751-42c4-951d-6008210024fa.png)

However, if you :

```
ddev restart
ddev drush cex -y
```

You will notice a diff is created

![image](https://user-images.githubusercontent.com/125707/192114543-67e0cb20-6bde-4a18-9177-cdf26fa71dca.png)

PR #316, opened 2022-09-24
Upgrade config ignore to latest beta

amitaibu:
fixes https://github.com/Gizra/drupal-starter/issues/315

Config now properly ignored

```
ddev drush cex -y
 [notice] The active configuration is identical to the configuration in the export directory (../config/sync).
../config/sync
```


bboro (approved):
> The behavior is different from 2.x because ignoring also happens on export.

:+1:

Issue #320, opened 2022-09-28
Fix webprofiler issue when saving fields


mariano-dagostino:
https://gitlab.com/drupalspoons/webprofiler/-/issues/43

PR #321, opened 2022-09-28
Patch webprofiler to allow saving fields

mariano-dagostino:
#320


composer.json
```diff
@@ -103,6 +103,9 @@
             },
             "drupal/core": {
               "d.o #3224254": "https://www.drupal.org/files/issues/2021-07-19/drupal-claro_dropdown-3224254-2.patch"
+            },
+            "drupal/devel": {
+              "d.o #43": "patches/419-Webprofiler-related-crash-in-core-9.4.2.patch"
```
amitaibu, line 108: Let's follow other patches, and use the d.o URL, rather than have it locally

amitaibu (approved):
Thanks - one comment, and MWG

composer.json
mariano-dagostino: @amitaibu devel module does not uses drupal.org and issue tracking is done at https://gitlab.com/drupalspoons/devel

mariano-dagostino: ah. I see your point...

amitaibu: GitHub allows creating patch files by appending `.patch` to the PR - for example https://github.com/Gizra/drupal-starter/pull/321.patch . 
Gitlab likely has something similar

mariano-dagostino: @amitaibu Yes, but there is no open PR for that issue, only a patch attached in a comment.

PR #323, opened 2022-09-29
Removed outdated robo watch task

dipakmdhrm:
#324

### DIFF
--- a/RoboFile.php
+++ b/RoboFile.php
@@ -163,56 +163,6 @@ class RoboFile extends Tasks {
   }
 
   /**
-   * Directories that should be watched for the theme.
-   *
-   * @return array
-   *   List of directories.
-   */
-  protected function monitoredThemeDirectories(): array {
-    return [
-      self::THEME_BASE . '/src',
-      self::THEME_BASE . '/templates',
-    ];
-  }
-
-  /**
-   * Watch the theme and compile on change (optimized).
-   */
-  public function themeWatch(): void {
-    $this->say('Compiling and watching (optimized).');
-    $this->doThemeCompile(TRUE);
-    $watch = $this->taskWatch();
-    foreach ($this->monitoredThemeDirectories() as $directory) {
-      $watch->monitor(
-        $directory,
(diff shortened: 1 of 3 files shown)


RoboFile.php
```diff
@@ -180,16 +181,17 @@ protected function monitoredThemeDirectories(): array {
   public function themeWatch(): void {
     $this->say('Compiling and watching (optimized).');
     $this->doThemeCompile(TRUE);
+    $watch = $this->taskWatch();
     foreach ($this->monitoredThemeDirectories() as $directory) {
-      $this->taskWatch()
```
dipakmdhrm, line 184: Reference from here: https://robo.li/tasks/Base/#watch

amitaibu:
Thanks. Can you please give https://github.com/Gizra/drupal-starter#theme-development an update. It's confusing what needs to be executed - we need to highlight the `ddev theme:watch`

dipakmdhrm:
@amitaibu While updating the README, I noticed that `ddev theme:watch` doesn't watch or compile js files or images. Is that intentional?

amitaibu:
> Is that intentional?

It's not. Are you up tp try and fix that? :)

dipakmdhrm:
> It's not. Are you up tp try and fix that? :)

Sure. I'll give it a try

dipakmdhrm:
@amitaibu Robo has (or 'had' after I removed it in last commit) it's own watch tasks, which looks like can't be combined with tailwind's css tasks. _Are we open to use a js bundler (webpack/gulp or others) for js & image build?_

What I plan to do is:
- Keep the existing robo image build task
- Keep the existing postcss css build task
- Create the new js build task
- Use the bundler to combine these three in a compile & a watch task.

amitaibu:
> Are we open to use a js bundler

No, tried a few in the past (Parcel and gulp). They always require more effort than we want. And Robo I believe should give us that (copy, optimize images, optimize JS)

dipakmdhrm:
Closing this for #330

PR #325, opened 2022-09-29
Update Drupal core and contrib modules and fix devel webprofiler patch

bboro:


bboro:
Note: Moved devel to dev dependencies

mariano-dagostino (approved):

bboro:
@amitaibu FYI ^

Issue #299, opened 2022-09-07
Add Test to confirm all Entity reference fields use Select2


amitaibu:

amitaibu:
Done in https://github.com/Gizra/drupal-starter/pull/342/files#diff-96631566c8a8c374694328b33bc56400fa8d04ae36bf9ea8b24e53e111b354afR52

Instead of tests, we make it the default widget.

PR #326, opened 2022-09-29
Add Test to confirm all Entity reference fields use Select2 

amitaibu:
#299

### DIFF
--- a/web/modules/custom/server_general/tests/src/ExistingSite/ServerGeneralEntityReferecesSelect2Test.php
+++ b/web/modules/custom/server_general/tests/src/ExistingSite/ServerGeneralEntityReferecesSelect2Test.php
@@ -1,0 +1,44 @@
+<?php
+
+namespace Drupal\Tests\server_general\ExistingSite;
+
+use weitzman\DrupalTestTraits\ExistingSiteBase;
+
+/**
+ * Tests entity references.
+ */
+class ServerGeneralEntityReferecesSelect2Test extends ExistingSiteBase {
+
+  /**
+   * Checks all entity references referencing nodes use the select2 widget.
+   */
+  public function testSelect2Widget() {
+    $em = \Drupal::entityTypeManager();
+    $field_configs = $em->getStorage('field_config')->loadMultiple();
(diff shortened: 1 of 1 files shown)


web/modules/custom/server_general/tests/src/ExistingSite/ServerGeneralEntityReferecesSelect2Test.php
```diff
@@ -0,0 +1,44 @@
...
+
+use weitzman\DrupalTestTraits\ExistingSiteBase;
+
+/**
+ * Tests entity references.
+ */
+class ServerGeneralEntityReferecesSelect2Test extends ExistingSiteBase {
+
+  /**
+   * Checks all entity references referencing nodes use the select2 widget.
```
amitaibu, line 13:
```suggestion
   * Checks all entity references are using the select2 widget.
```

web/modules/custom/server_general/tests/src/ExistingSite/ServerGeneralEntityReferecesSelect2Test.php
```diff
@@ -0,0 +1,44 @@
...
+    foreach ($field_configs as $id => $field_config) {
+      if ($field_config->getType() !== 'entity_reference') {
+        continue;
+      }
+
+      $storage_definition = $field_config->getFieldStorageDefinition();
+      $referencing = $storage_definition->getSetting('target_type');
+      if ($referencing != 'node') {
+        continue;
+      }
```
amitaibu, line 29: Let's always check that

```suggestion
      if ($referencing != 'node') {
        continue;
      }
```

amitaibu (commented):
While reviewing this, I thought -- maybe instead of test, we can change the default widget of ER to be select2?

amitaibu:
> maybe instead of test, we can change the default widget of ER to be select2?

@dipakmdhrm let's check this option ^^

amitaibu:
Done in https://github.com/Gizra/drupal-starter/pull/342/files#diff-96631566c8a8c374694328b33bc56400fa8d04ae36bf9ea8b24e53e111b354afR52

Issue #289, opened 2022-08-30
Add `assertFieldIsRequired` tests to Content types and Paragraph types [2h]


amitaibu:
To assert the fields and bundle using `drupal/drupal_test_assertions`

amitaibu:
@mariano-dagostino can you please tackle this - we need it in other projects :smile:

amitaibu:
@dipakmdhrm this one is yours now :smile:

amitaibu:
@dipakmdhrm 

Thanks, I think worth a small follow up:

`ServerGeneralContentTypeTestBase` => `ServerGeneralEntityTestBase`
`ServerGeneralContentTypeNews` => `ServerGeneralNodeNews`

PR #327, opened 2022-09-30
Add assertFieldIsRequired tests to Content types and Paragraph types

dipakmdhrm:
#289 

- Add `assertFieldIsRequired` & 'assertFieldIsNotRequired' tests to Content types and Paragraph types.

**TB:1.7/2h**

### DIFF
--- a/web/modules/custom/server_general/tests/src/ExistingSite/RequiredAndOptionalFieldTestInterface.php
+++ b/web/modules/custom/server_general/tests/src/ExistingSite/RequiredAndOptionalFieldTestInterface.php
@@ -26,7 +26,7 @@ interface RequiredAndOptionalFieldTestInterface {
   /**
    * The required fields for entity bundle.
    *
-   * @return array
+   * @return string[]
    *   Array of required field names.
    */
   public function getRequiredFields() : array;
@@ -34,7 +34,7 @@ interface RequiredAndOptionalFieldTestInterface {
   /**
    * The optional fields for entity bundle.
    *
-   * @return array
+   * @return string[]
    *   Array of optional field names.
    */
   public function getOptionalFields() : array;
--- a/web/modules/custom/server_general/tests/src/ExistingSite/RequiredAndOptionalFieldTestInterface.php
+++ b/web/modules/custom/server_general/tests/src/ExistingSite/RequiredAndOptionalFieldTestInterface.php
@@ -2,10 +2,13 @@
 
 namespace Drupal\Tests\server_general\ExistingSite;
 
+/**
+ * Interface defining required function for required/optional field tests.
+ */
 interface RequiredAndOptionalFieldTestInterface {
 
   /**
-   *  The entity type to assert it exists.
+   * The entity type to assert it exists.
    *
    * @return string
    *   An entity type name.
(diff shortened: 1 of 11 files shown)


web/modules/custom/server_general/tests/src/ExistingSite/ServerGeneralEntityTypeTestBase.php
```diff
@@ -0,0 +1,40 @@
...
+ */
+abstract class ServerGeneralEntityTypeTestBase extends ExistingSiteBase {
+
+  use EntityTrait;
+  use FieldsTrait;
+
+  const ENTITY_TYPE = '';
+  const ENTITY_BUNDLE = '';
+  const REQUIRED_FIELDS = [];
+  const OPTIONAL_FIELDS = [];
```
mariano-dagostino, line 20: I think we could either, document the types of each const here.. Or maybe better define an Interface with four methods and they respective return types.

Something like

```
interface RequiredAndOptionalFieldTestInterface {

  /**
   *  The entity type to assert it exists.
   *
   * @return string
   *   An entity type name.
   */
  public function entityType() : string;

  ...

}

So the compiler will let us know there is something off in new tests.

**Review by mariano-dagostino — CHANGES_REQUESTED** [review]

File: web/modules/custom/server_general/tests/src/ExistingSite/ServerGeneralEntityTypeTestBase.php (hunk 1 above)
Comment (dipakmdhrm) [reply web/modules/custom/server_general/tests/src/ExistingSite/ServerGeneralEntityTypeTestBase.php]: That's a great suggestion. Thank you!
It's done.

**Comment by dipakmdhrm** [comment]
### Update: 

- Added RequiredAndOptionalFieldTestInterface interface

**TB:1.9/2h**

File: web/modules/custom/server_general/tests/src/ExistingSite/RequiredAndOptionalFieldTestInterface.php [hunk 2]
```diff
@@ -0,0 +1,42 @@
...
+   *
+   * @return string
+   *   A bundle name.
+   */
+  public function getEntityBundle() : string;
+
+  /**
+   * The required fields for entity bundle.
+   *
+   * @return array
```
mariano-dagostino, line 29: @dipakmdhrm would be great to use `[]string` here. I think phpstan can detect this as well.

web/modules/custom/server_general/tests/src/ExistingSite/RequiredAndOptionalFieldTestInterface.php
```diff
@@ -0,0 +1,42 @@
...
+   *   Array of required field names.
+   */
+  public function getRequiredFields() : array;
+
+  /**
+   * The optional fields for entity bundle.
+   *
+   * @return array
```
mariano-dagostino, line 37: Same here

mariano-dagostino (commented):
@dipakmdhrm Two more changes and MWG.

web/modules/custom/server_general/tests/src/ExistingSite/ServerGeneralContentTypeNews.php
```diff
@@ -0,0 +1,36 @@
+<?php
+
+namespace Drupal\Tests\server_general\ExistingSite;
+
+/**
+ * Test 'news' content type.
+ */
+class ServerGeneralContentTypeNews extends ServerGeneralContentTypeTestBase {
```
bboro, line 8: Test classes need to end with "Test" otherwise they don't run: https://app.travis-ci.com/github/Gizra/drupal-starter/jobs/584404313 look for `ServerGeneralContentTypeNews` it's not listed in the list of tests ran

PR #328, opened 2022-09-30
Install and enable focal_point

bboro:


amitaibu:
Needs a re-roll.
Do we have the focal point on the media items as-well? That is, when a user adds a News item, can they set a focal point?

bboro:
@amitaibu now they can.

Updated cards to also use an image style with focal point crop

Before|After
-|-
image|image

TB: +0.5h

PR #329, opened 2022-10-03
Simplify DDEV install

AronNovak:


ci-scripts/install_ddev.sh
```diff
@@ -4,12 +4,6 @@ set -e
 echo "Logging into Docker Hub"
 docker login --password "$DOCKER_PASSWORD" --username amitaibu
 
-echo "Install mkcert."
```
amitaibu, line 7: Isn't it needed anymore for local SSL?

AronNovak: It is needed, but DDEV install script takes care of it. It's better if it self-manages the install (versions, etc).

AronNovak: https://github.com/drud/ddev/blob/81ec233273208fcad39c37759abdc36ceadf8bc1/scripts/install_ddev.sh#L197 and so on

amitaibu: Oh, seems it's new. I think in the past also for Linux we had to annually install mckert.

AronNovak: Yep, anyways as Travis proves, PHPUnit is satisfied with the current conditions :smile_cat:

Issue #324, opened 2022-09-29
Add theme template directory to list of directories for watch task.


dipakmdhrm:
Theme template directory is not a part of watch task at the moment. 

If a developer is running the watch task, and makes changes to a template file, say add a tailwind utility class, they'll either have to restart the watch task, or make some empty changes in the src directory.

Adding template directory to watch task will avoid this.

amitaibu:
> say add a tailwind utility class

It does work for me as we watch those [files](https://github.com/Gizra/drupal-starter/blob/dc047ae29b87df110c0fb0d553d5b99816f2c95b/web/themes/custom/server_theme/tailwind.config.js#L16-L17). Have you tried with latest Drupal-starter?

dipakmdhrm:
> > say add a tailwind utility class
> 
> It does work for me as we watch those [files](https://github.com/Gizra/drupal-starter/blob/dc047ae29b87df110c0fb0d553d5b99816f2c95b/web/themes/custom/server_theme/tailwind.config.js#L16-L17). Have you tried with latest Drupal-starter?

That's my bad. I've been using the older watch task using robo `ddev robo theme:watch`. Closing this PR & issue.

amitaibu:
Let's remove the old one (from Robo and docs, if they are still there)

dipakmdhrm:
#323

amitaibu:
[Reverted](https://github.com/Gizra/drupal-starter/pull/346) - this needs some more work. Compiling CSS should work as fast as we have it now

PR #330, opened 2022-10-03
Update theme watch to include js & images.

dipakmdhrm:
#324 

- Updated theme watch to include js & images.
- Updated README


AronNovak (approved):
It makes sense, thanks! Indeed it's a frustration when custom code doesn't get styling due to the fact that we don't watch that.

The only thing I'd advise is to not introduce redundancy for documentation. On the long run one of the `README.md`s will become outdated. Let's keep the existing README file, and if you prefer, let's just refer to the top README from the theme README, instead of duplicating part of its content.

dipakmdhrm:
> It makes sense, thanks! Indeed it's a frustration when custom code doesn't get styling due to the fact that we don't watch that.
> 
> The only thing I'd advise is to not introduce redundancy for documentation. On the long run one of the `README.md`s will become outdated. Let's keep the existing README file, and if you prefer, let's just refer to the top README from the theme README, instead of duplicating part of its content.

Done.

web/themes/custom/server_theme/README.md
```diff
@@ -1,5 +1,7 @@
 # Theme Development
 
+To learn about theme directory structure and commands to compile and watch
```
amitaibu, line 3: Let's remove this file - and move all info to the main README.md

dipakmdhrm: Done.

PR #338, opened 2022-10-10
Set the carousel direction based on the current language

amitaibu:
| Without PR | With PR |
| -- | -- |
| image | image |


bboro (approved):
:+1:

PR #339, opened 2022-10-10
Add Arabic and Spanish languages

amitaibu:
image

We'll have follow up PRs to add related functionality (e.g. Language switcher and Search)

### DIFF
--- a/config/sync/core.entity_form_display.block_content.basic.default.yml
+++ b/config/sync/core.entity_form_display.block_content.basic.default.yml
@@ -32,4 +32,11 @@ content:
       size: 60
       placeholder: ''
     third_party_settings: {  }
+  langcode:
+    type: language_select
+    weight: 2
+    region: content
+    settings:
+      include_locked: true
+    third_party_settings: {  }
 hidden: {  }
--- a/config/sync/core.entity_form_display.block_content.basic.default.yml
+++ b/config/sync/core.entity_form_display.block_content.basic.default.yml
@@ -37,6 +37,6 @@ content:
     weight: 2
     region: content
     settings:
-      include_locked: true
+      include_locked: false
     third_party_settings: {  }
 hidden: {  }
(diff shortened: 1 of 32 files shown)


config/sync/core.entity_form_display.block_content.basic.default.yml
```diff
@@ -32,4 +32,11 @@ content:
       size: 60
       placeholder: ''
     third_party_settings: {  }
+  langcode:
+    type: language_select
+    weight: 2
+    region: content
+    settings:
+      include_locked: true
```
bboro, line 40: It would be good if we could do a search&replace for `include_locked: true` and change it to `false`. This remove the `Not defined` and `Not applicable` languages from the dropdown.

bboro: We could also add a test to ContentTypeTestBase to ensure that devs remember to do this for new node types. **[For example like this](https://github.com/Gizra/WMO/blob/main/web/modules/custom/server_general/tests/src/ExistingSite/ContentTypeTestBase.php#L47)**

Edit: Looks like we're hiding the Language selector for most entities, so this is probably not applicable.

config/sync/language.mappings.yml
```diff
@@ -0,0 +1,13 @@
...
+  'no': nb
+  pt: pt-pt
+  zh: zh-hans
+  zh-tw: zh-hant
+  zh-hk: zh-hant
+  zh-mo: zh-hant
+  zh-cht: zh-hant
+  zh-cn: zh-hans
+  zh-sg: zh-hans
+  zh-chs: zh-hans
```
bboro, line 13: Why do we have Norwegian, Portuguese and Chinese mappings here?

bboro (approved):

config/sync/core.entity_form_display.block_content.basic.default.yml
amitaibu: > Looks like we're hiding the Language selector for most entities, so this is probably not applicable.

I think we actually need to show it. So one could fill in Spanish, but with English user interface (which is likely, but kind of projects we deal with, often they copy/ paste from another source)

amitaibu:
Closed in favor of https://github.com/Gizra/drupal-starter/pull/342

PR #347, opened 2022-10-15
Overhaul design to take advantage of vertical spacing wrapper

amitaibu:
- Add and fix missing Social share buttons
- Remove the top/ bottom margin from the first/ last element of a prose.
- Remove padding/ margins from elements in favor of `flex flex-col gap-y-5`. This opens the door to more composition and closes the door on un-wanted variations, that would cause inconsistencies and harder development. So now we can wrap elements on a PEVB with `$this->wrapContainerVerticalSpacing`. See `NodeNews` PEVB which got a new design with a sidebar.
Also `NodeLandingPage` is taking advantage fo the new `wrapContainerVerticalSpacing`.

| Landing page | News page |
| -- | -- |
| Screenshot 2022-10-15 at 17-33-48 Homepage Drupal Starter | Screenshot 2022-10-15 at 17-35-11 Worrying Study Finds Today's Tutorial Levels Not Adequately Preparing Players For Challenges Of World 8 Drupal Starter  |


bboro (approved):
Looks fine to me but I have some ideas/suggestions:

1. For maintainability sake, would it be better to just create 1 master twig, with some config variables like `spacing = 'tiny'/'big'`. We can still have `server-theme-container-vertical-spacing-big.html.twig`, but its contents would just be `{% @extends .... %} {% set spacing = 'big' %}`

1. I noticed on a lot of our designs we have custom spacings for mobile/tablet/desktop breakpoints. Should we perhaps include some mobile/tablet/desktop spacings into these spacing wrappers, for example `gap-y-5 md:gap-y-8 lg:gap-y-10`. Feels like it would be good to agree on the standard spacings with @nora-novak.

amitaibu:
> For maintainability sake, would it be better to just create 1 master twig

I think that it's for such a small twigs, we can go without extend, as it won't save much (and just add another base twig)`

>  for example gap-y-5 md:gap-y-8 lg:gap-y-10

Good catch, I've actually added it to another project. I'll copy it over here.

PR #362, opened 2022-11-14
Pantheon Advanced Page Cace

AronNovak:
> Just by turning on this module your Drupal site will start emitting the HTTP headers necessary to make the Pantheon Global CDN aware of data underlying the response. Then, when the underlying data changes (nodes and taxonomy terms are updated, user permissions changed) this module will clear only the relevant pages from the edge cache.

Not sure how we missed this must-have module so far.


amitaibu (approved):
:ok_hand:

PR #373, opened 2022-11-24
Add Embed media to CK editor

amitaibu:
- Added Video Media
- Enabled Media embed module
- Added to CK editor for both Basic HTML and Full HTML.

Couple of issues I see:

1. The Embed button doesn't work for the Full HTML
2. When adding an Embed, it also adds the Caption to the News body. Then, if we re-edit the node and save, the captions are being multiplied.

| First  save | Second save |
| -- | -- |
| image | image |


amitaibu:
> The Embed button doesn't work for the Full HTML

Fixed. `<drupal-entity>`'s attributes were missing from the config

amitaibu:
Switched to core's Media embed, which looks nicer and behaves better. We can save multiple times, and no duplicate caption. However, once we click the `Edit Media` and save on the modal, then we get the junk HTML added to the body field.

image

image

@bboro can you please take this from here?

amitaibu:
- Using oembed formater
- Added a generic `::buildCaption` that can be used on `Image` and `Video` media types
- Added a trait with `::buildVideo` to be used on PEVB and the style guide

image

We still have an issue with the embed from CKeditor, so once I've added the `MediaVideo` pevb, once cannot embed a video

amitaibu:
https://github.com/Gizra/drupal-starter/pull/373/commits/d99cf801ecd21ff9a65dfc9577c7cc3f6469e2a7 should fix the duplication issue. `figcaption` was recognized by CK editor, as an element that requires own editing. Luckily there was some visual indication that helped me find it

| `<div>` | `<figcaption>` |
| -- | -- |
| image | image |

So I've replaced `figcaption` with `div`, and CKeditor is happy now

amitaibu:
- Removed the Alignment (left, center, right) when embedding media, as we're always placing it in own row
- Prose is now max width of 768px (slightly higher than the 706px), so the body images align with the featured image
- Added `wrapRoundedCornersBig` to wrap elements with `lg` rounded corner
- Media image no longer up-scales items, so user can add smaller images as well

Screenshot 2022-11-27 at 11-18-07 Try Drupal Starter

config/sync/responsive_image.styles.prose_image.yml
```diff
@@ -0,0 +1,13 @@
...
+langcode: en
+status: true
+dependencies:
+  config:
+    - image.style.prose_image
+  theme:
+    - server_theme
+id: prose_image
+label: 'Prose image'
+image_style_mappings: {  }
```
bboro, line 11: Shall we add a 2x for retina displays and put the mappings here

web/modules/custom/server_general/src/Plugin/EntityViewBuilder/MediaImage.php
```diff
@@ -20,41 +21,84 @@
...
+   *
+   * @param array $build
+   *   The build array.
+   * @param \Drupal\media\MediaInterface $entity
+   *   The entity.
+   *
+   * @return array
+   *   The render array.
+   */
+  public function buildFull(array $build, MediaInterface $entity): array {
```
bboro, line 47: We usually use `embed` view mode on media (at least on other projects) to output wysiwyg embeds. Plus this is all tied to the existing migrate plugins too:

```php
  /**
   * Creates a DOM element representing an embed media on the destination.
   *
   * @param \DOMDocument $dom
   *   The \DOMDocument in which the embed \DOMElement is being created.
   * @param string $media_uuid
   *   The UUID of the media which should be represented by the new embed tag.
   *
   * @return \DOMElement
   *   The new embed tag as a writable \DOMElement.
   */
  protected function createEmbedElement(\DOMDocument $dom, $media_uuid): \DOMElement {
    $element = $dom->createElement('drupal-entity');
    $element->setAttribute('data-entity-type', 'media');
    $element->setAttribute('data-entity-uuid', $media_uuid);
    $element->setAttribute('data-embed-button', 'media');
    $element->setAttribute('data-entity-embed-display', 'view_mode:media.embed');
    $element->setAttribute('data-entity-embed-display-settings', '');

    return $element;
  }
```

Plus this also means we should use `entity_embed` (`<drupal-entity`) instead of core embed (`<drupal-media), or update the migration plugins.

web/modules/custom/server_general/src/Plugin/EntityViewBuilder/MediaVideo.php
```diff
@@ -0,0 +1,73 @@
...
+   *
+   * @param array $build
+   *   The build array.
+   * @param \Drupal\media\MediaInterface $entity
+   *   The entity.
+   *
+   * @return array
+   *   The render array.
+   */
+  public function buildFull(array $build, MediaInterface $entity): array {
```
bboro, line 58: Also `embed` view mode here? But I suppose full and embed view modes can output same content here. So maybe

```php
public function buildEmbed($build, $entity) {
  return $this->buildFull($build, $entity);
}
```

web/themes/custom/server_theme/templates/server-theme-media--image.html.twig
```diff
@@ -0,0 +1,7 @@
+<div class="flex flex-col gap-y-2 md:gap-y-3 mb-5">
+  <figure class=overflow-hidden">
```
bboro, line 2:
```suggestion
  <figure class="overflow-hidden">
```

web/modules/custom/server_general/src/Plugin/EntityViewBuilder/MediaImage.php
amitaibu: Is there an advantage to `entity_embed`. From what I saw, core's embed is nicer:

- Major; Part of core
- We really intend to embed only `Media`, while entity-embed seems to be also for embedding other entities (e.g. the `Node` embed it creates). So there's probably a bunch of code there we don't need.
- Minor; The modal is better

> , or update the migration plugins.

What's the approximate effort to do that?

web/modules/custom/server_general/src/Plugin/EntityViewBuilder/MediaVideo.php
amitaibu: tbh, I went with `Full`just because it's out of the box, and didn't need to declare another view mode.

web/modules/custom/server_general/src/Plugin/EntityViewBuilder/MediaImage.php
bboro: I think we just need to map drupal-entity's data-* attributes to drupal-media's and run some test migrations. I suppose we can add some `<img` or drupal 7 style embeds (`[[{"fid":"123"...`) to the CSV migrations and run the processors on the body field to check they're working too.

So around 2-3h maybe?

amitaibu: @bboro 👍 

Do you see any downside to going this route?

web/modules/custom/server_general/src/Plugin/EntityViewBuilder/MediaVideo.php
bboro: I think it'll be good to have a dedicated `embed` view mode for media since we're adding the embed functionality (plus it's also tied to migrations: `$element->setAttribute('data-entity-embed-display', 'view_mode:media.embed');`. 

Otherwise we'll almost certainly need to change this on new projects each time, if they need a full media view page.

web/modules/custom/server_general/src/Plugin/EntityViewBuilder/MediaImage.php
bboro: @amitaibu I'm not familiar with core embed plugin so I'm not really sure. 

I think originally it was switched from Entity Embed -> Core Embed because of the caption double output issue right? Which was due to something `<figcaption` and unrelated to entity_embed. 

I guess ultimately it depends on how customizable the core one is, in terms of media types allowed, view modes allowed etc. So I'll leave it up to you :)

amitaibu:
@bboro I've addressed your comments. Can you take [the migration part](https://github.com/Gizra/drupal-starter/pull/373#discussion_r1033496591) please?

bboro (approved):

bboro:
@amitaibu updated migration plugins

PR #379, opened 2022-12-01
Reorder languages in mobile language selector.

balagan73:
image

Reorder languages in mobile language selector, and put the current language on top.


amitaibu (approved):
Thanks

PR #381, opened 2022-12-05
Improve array filtering

amitaibu:


web/modules/custom/server_general/src/ElementWrapTrait.php
```diff
@@ -226,4 +226,25 @@ protected function wrapRoundedCornersBig(array $element): array {
...
+   * @return array
+   *   The filtered render array.
+   */
+  protected function filterEmptyElements(array $element): array {
+    if (count($element) <= 1) {
+      // No change.
+      return $element;
+    }
+
+    return array_filter($element);
```
bboro, line 247: This still filters out 

```php
$element = [
  '#theme' => 'server_theme_x',
  '#some_count' => '0',
  '#is_something' => FALSE,
];
```

because count $element here would return 3 so it will pass it on to `array_filter`. Perhaps we should do something like this:

```suggestion
    if (count(\Drupal\Core\Render\Element::properties($element))) {
      // Element has top level properties beginning with #.
      // Do not filter.
      return $element;
    }

    return array_filter($element);
```

amitaibu: Thanks. I'll write a small test to validate

web/modules/custom/server_general/tests/src/ExistingSite/ServerGeneralElementWrapTest.php
```diff
@@ -0,0 +1,72 @@
...
+    $element = [
+      '#foo' => FALSE,
+      0 => [],
+    ];
+    $expected = [
+      '#foo' => FALSE,
+      0 => [],
+    ];
+    $result = $this->filterEmptyElements($element);
+    $this->assertEquals($expected, $result);
```
bboro, line 49: :+1: cool simple test

bboro (approved):

PR #384, opened 2022-12-08
The Fillmurray.com service is down

balagan73:
Replaced the fillmurray.com url to create person images on the style guide.


web/modules/custom/server_style_guide/src/Controller/StyleGuideController.php
```diff
@@ -430,7 +430,7 @@ protected function getPlaceholderImage(int $width, int $height, string $id = '',
    *   URL with placeholder.
    */
   protected function getPlaceholderPersonImage(int $width, int $height, string $text = NULL) {
-    return "https://www.fillmurray.com/{$width}/{$height}" . (!empty($text) ? '?text=' . $text : NULL);
+    return "https://placebeard.it/{$width}/{$height}" . (!empty($text) ? '?text=' . $text : NULL);
```
amitaibu, line 433: Let's give https://pravatar.cc/ a try - so we don't have only pics of dudes :)

amitaibu:
Added Profile picture to style guide

image

PR #386, opened 2022-12-13
Ensure rounded corners fit to their content

bboro:
This ensures the rounded-corners div wrapper nicely wraps the image, otherwise the div stretches full width, and only the top/bottom left corners are rounded.

No `w-fit`|With `w-fit`
-|-
image|image

Difference: Top-right, bottom-right corners are rounded on a portrait image


amitaibu (approved):
I love the choice of image :smile:

PR #390, opened 2022-12-16
Prepare menu links recursively.

balagan73:
I think we might want to port this from another project

Single level menu links still display properly:

image


bboro (approved):
This LGTM

PR #397, opened 2022-12-22
Autodeployment fixes

### LINKED ISSUE #395: Autodeployment issues and improvements
## All default values used by robo end with a dot

The use of `Default: something.` as part of method descriptions breaks all expected values: For example:

```
  /**
   * Install the site on specific env on Pantheon from scratch.
   *
   * Running this command via `ddev` will require terminus login inside ddev:
   * `ddev auth ssh`.
   *
   * @param string $env
   *   The environment to install. Default: qa.
   *
   * @throws \Exception
   */
  public function deployPantheonInstallEnv(string $env = 'qa'): void {
```

You think the default value for env is `qa` but when no value is specified, the default value is `qa.` which breaks all the logic inside.

## Script ci-scripts/prepare_deploy.sh needs to run terminus auth:login.

The [pantheon integration for ddev](https://ddev.readthedocs.io/en/latest/users/providers/pantheon/) seems to use the terminus token only when running ddev pull, and ddev push... But we also need to login into terminus when preparing the deployment. An extra `ddev . terminus auth:login --machine-token="$TERMINUS_TOKEN"` is required in prepare_deploy.sh

## There is no way to force a deploy if there are no code changes

Sometimes deployment fail because some content is present in the QA environment and you need to manually delete them. If you attempt to run the same deployment again to finish it properly, the automated process will indicate there is no code to commit, and stop the deployment there:

```
 Nothing to commit, working tree clean
The command "travis_retry ddev robo deploy:pantheon --no-interaction qa" exited with 0.
```

## Config autodeploy is not capable of regenerating secrets

In case of need regenerating secrets it is not possible to run deploy-config-autodeploy cleanly, because all tokens has been already replaced. Also it complains about some existing files in the source code.

Maybe is worth to have a .travis.yml.template and use it as base when running this command.

## deploy:pantheon-install-env ask for confirmation when uninstalling the migrate module.

There is a missing ` -y` that needs to be added to the drush command.

mariano-dagostino:
#395

### DIFF
--- a/.travis.template.yml
+++ b/.travis.template.yml
@@ -1,0 +1,64 @@
+version: ~> 1.0
+sudo: required
+language: php
+dist: focal
+php:
+  - 8.0
+
+services:
+  - docker
+
+stages:
+  - lint
+  - test
+  - deploy
+
+env:
+  global:
+    - PANTHEON_GIT_URL="{{ PANTHEON_GIT_URL }}"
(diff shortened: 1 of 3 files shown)


.travis.template.yml
```diff
@@ -0,0 +1,64 @@
...
+
+stages:
+  - lint
+  - test
+  - deploy
+
+env:
+  global:
+    - PANTHEON_GIT_URL="{{ PANTHEON_GIT_URL }}"
+    - secure: bE9Qwgm4OnzsHCtE4WBiQWNYRCoG0pO2wC0tY2bu0RVJ+TJekZ2UBC2XAr6dsoFCAZE1nYN6VnRMYuUqXThRjlaLEY96L2I2XD3m7iZrOlDVVEWP0vnOZD892x4k/S/wpQFXYaQHSyikyyFfL5m3V0bIWjoweyON293q5494WKRnVrD5+uYGMdSLQmwF8lm/L/q3hlLpm8Ju63IM4WHAk0YY8Z/ASjCsKk/QGJTePU5OSjKO+NFLU49LRMPXjbWxWiz9leWNy0NP8St+7ONW/QX9QPRxiOoBGK+gzz3U0zGmAbF3mWNfTC1KRIsiA9vVI36pYAfA2n2Bv4Jz31ktUkTOCt6nvkpL4/4LcYcp8uymM0wW2+coaWlLBxRIrjz80XgtAr3ruy6ArC6BddOefYCxsw13Z2LseG88FmRy4VX/ypIj7h38wdhQZqgj6WsEmQR2tHO3mlXcd+ZL/LdFA181uQhOPuPWn5UYEgC3Rth2N3n2OXguzFizFA/XBbPmDlG/KUPcJX4suUzhawvaMS79XvUIVEYID8InoMzQs3VtqlGeOqYHbz4rjn7inpQgdLw6OdIssk6pp0nwqugiCPrMlLw3G3yiUwNQZ9jXR0UjhOXR40KIGbGwg0xbC/xtdQjM+76VEtnJmpQQcDxZ/ZmEbUK8xdOuss84LFJnVKw=
```
mariano-dagostino, line 19: @AronNovak Not sure if we can delete this from the template... Probably we can

AronNovak: Yep, we can safely delete this, it's for Docker Hub auth, that we do not use for client projects.

AronNovak (approved):

PR #399, opened 2022-12-27
More wrappers and compose cards

amitaibu:
This is an experiment at reducing the custom cards in favor of composing twigs.

image


amitaibu:
First theme to go in favor of a PEVB solution is the search results card

image

amitaibu:
More composed cards

Screenshot 2022-12-29 at 01-27-30 Style Guide Drupal Starter

amitaibu:
- Added a centered card (can be used for logos of donors, or profile info)
- Improved narrow container to be in the middle
- Refine featured cards design

image

amitaibu:
I've created a more complex example for the Profile pic, to show how to do center alignment (`$this->wrapContainerVerticalSpacing($inner_elements, 'center');`)

image

amitaibu:
@bboro ready for review. Notable changes in the PR:

- `server-theme-card--name-of-content-type` is now less likely to happen. Like on the page level, also cards are now built using existing elements on PEVB. The idea isn't that we try to avoid having twigs, rather that we try to have too specific twigs, which are very similar to other ones. So we encourage the structure of the card/ element to be done by twig. And for the internal elements we re-use and compose on PEVB.
- We have a `CardTrait` and `ElementTrait`, which are wired both to the Style guide and to PEVB
- `buildButton`'s $url type hint changed from `string` to `Url`
- Re-order Style guide to make more sense
- Remove Views paragraph in favor of specific ones (e.g. `News teasers`)

## Style guide

| 639 px | 767px | 1279 px | RTL |
| -- | -- | -- | -- |
| drupal-starter ddev site_4443_style-guide (4) | drupal-starter ddev site_4443_style-guide (5) | drupal-starter ddev site_4443_style-guide (6) | drupal-starter ddev site_4443_ar_style-guide |

## Homepage

drupal-starter ddev site_4443_

web/modules/custom/server_general/src/ButtonTrait.php
```diff
@@ -58,7 +59,7 @@ protected function buildLinkButton(ContentEntityInterface $entity, string $field
     $links = $entity->{$field_name}->getValue();
```
bboro, line 59: We have `getLinkFieldValue` in PEVB Buildfieldtrait, which a) checks if the field has data b) checks access to the links and c) provides an already built `Drupal\Core\Url` object. Perhaps be better to use that?

web/modules/custom/server_general/src/ButtonTrait.php
```diff
@@ -81,7 +82,8 @@ protected function buildFileButton(ContentEntityInterface $entity, string $field
     $value = $entity->get($field_name)->getValue();
     $title = !empty($value[0]['description']) ? $value[0]['description'] : $this->t('Download');
 
-    return $this->buildButton($file->createFileUrl(), $title);
+    $url = Url::fromUri($file->createFileUrl());
```
bboro, line 85:
```suggestion
    $url = Url::fromUri($file->getFileUri());
```

bboro (approved):
I mean it's fine I guess. but again this seems to take away more power from devs and designers and Twig in order to have common builders in PHP. What's actually the advantage of having multiple `wrapThisWithSomething` methods, versus multiple `server_theme_card__something`? Seems like we're going to have the latter anyway, even with this implementation.

amitaibu:
@bboro That is a fair point and the reason I've asked you to review it is that I assumed you'd ask it :smile: 

>  but again this seems to take away more power from devs and designers 

I don't think more than we already do with PEVB on the page level. For the designers however it will provide a strict framework, where things are more consistent.

I'm observing one of our internal projects. Everything that's page-related doesn't require any theming. All the elements are already there, and we re-use them.
But I still see works and inconsistencies around cards. We end up duplicating stuff, and can end easily with inconsistencies. Here's an example:


```twig

{# server-theme-card.html.twig #}
<a href="{{ url }}" class="block h-full">
  <article class="h-full flex flex-col rounded-lg shadow-md overflow-hidden">

    <figure class="child-object-cover w-full ">
      {{ image }}
    </figure>

    <div class="flex flex-col gap-4 md:gap-5 p-4 md:p-5">
      {% if label or date %}
        <div class="flex flex-col gap-4 md:gap-2">
          {{ label }}
          <span class="text-sm lg:text-lg">{{ date }}</span>
        </div>
      {% endif %}

      <h2 class="text-base md:text-lg lg:text-2xl font-semibold">{{ title }}</h2>
    </div>
  </article>
</a>
```

VS

```twig
{# server-theme-card--digital-resource.html.twig #}
<article class="h-full flex flex-col rounded-lg shadow-md overflow-hidden">
  <figure class="child-object-cover w-full ">
    {{ image }}
  </figure>

  <div class="flex flex-col gap-4 md:gap-5 p-4 md:p-5">
    <h2 class="text-base md:text-lg lg:text-2xl font-semibold">{{ title }}</h2>
    <div class="text-wmo-blue-gray-900 text-sm md:text-base lg:text-lg">
      {{ description }}
    </div>

    {% if url %}
      {% embed '@server_theme/server-theme-cta-button-round.html.twig' %}
        {% set is_external_url = url.isExternal() %}
        {% set button_type = 'dark' %}
      {% endembed %}
    {% endif %}
  </div>
</article>

```
You can see the later is a copy/ paste with some adaptation. It's easy to imagine how a fix in one will not reach the other.
Instead, for this similar case, we'd use `buildCardWithImage`, or similar.

> What's actually the advantage of having multiple wrapThisWithSomething methods, versus multiple server_theme_card__something?

We're still going to have a few `server_theme_card__something`, but they will not be content-type specific, as they are right now. For example, in the internal project we have this:

image

For me a twig with a content type name is becoming a code smell.

> Seems like we're going to have the latter anyway, even with this implementation.

Indeed, my goal isn't to artificially reduce twigs in favor of PHP. Layout is created with twig. Only internal text/ elements are then built the same we do for pages, by composition and consistent vertical spacing.

bboro:
Well if you put it that way :smile: it makes sense! I think we should perhaps try to give Nora a general overview of what we're doing here as well. But I guess she's already designing things in a way that re-uses a lot of components anyway, so perhaps not?

PR #403, opened 2023-01-06
Fix default body field machine name method argument.

mgurjanov:
Default value for $field argument in method buildProcessedText() is wrong. The @param comment has correct value. Instead of "field_body" it should default to "body".


amitaibu:
[Done](https://github.com/Gizra/drupal-starter/pull/405/files)

PR #404, opened 2023-01-10
Improve array filtering for element wrap.

iswariya:
If element has empty key, Element::Properties throws error: Uninitialized string offset: 0 in Drupal\Core\Render\Element::property(). To fix this issue, before passing it to Properties method, filtering out the empty keys.

### DIFF
--- a/web/modules/custom/server_general/tests/src/ExistingSite/ServerGeneralElementWrapTest.php
+++ b/web/modules/custom/server_general/tests/src/ExistingSite/ServerGeneralElementWrapTest.php
@@ -28,9 +28,11 @@ class ServerGeneralElementWrapTest extends ExistingSiteBase {
     // Non-nested array.
     $element = [
       '#foo' => FALSE,
+      '' => FALSE,
     ];
     $expected = [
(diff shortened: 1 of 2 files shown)


web/modules/custom/server_general/tests/src/ExistingSite/ServerGeneralElementWrapTest.php
```diff
@@ -67,6 +67,40 @@ public function testFilterEmptyElements() {
     ];
     $result = $this->filterEmptyElements($element);
     $this->assertEquals($expected, $result);
+
+    // Non-nested array with empty key.
```
amitaibu, line 71: Move this above `Mix of non-nested and nested array`

Then you can remove the duplicates lower ones. Instead, change the original ones to have the empty key check. For example:

```php
    // Mix of non-nested and nested array. As it has top level `#` we shouldn't
    // filter it at all.
    $element = [
      '#foo' => FALSE,
      0 => [],
      '' => FALSE,
    ];

bboro (approved):

PR #407, opened 2023-01-11
Throw exception when deployment fails to avoid proceeding and update travis clone depth

bboro:


.travis.yml
```diff
@@ -22,6 +22,12 @@ before_install:
   - export PATH="$HOME/.config/composer/vendor/bin:$PATH"
   - docker login --password "$DOCKER_PASSWORD" --username amitaibu
 
+git:
+  # By default, travis clones with depth 50 for performance, however this can
+  # cause issues with deployment as the full history is not available. So we
+  # disable depth specification and clone the full branch history.
+  depth: false
```
amitaibu, line 29: Let's see if we are able to improve the version detection without this change - as it might slow down travis.

@bboro Can you describe the scenario where a bigger commit history is needed?

bboro: Tag based deployments. See #devstuff in slack.

AronNovak: One option is to completely change strategy and use locking to make sure we never override history. This is to serialize the execution of the deploy jobs, one could wait until the other ends, or simply abort it this case. But it's a bit painful, as if the lock is stale, we can't deploy at all, and so on.

bboro: Alternative suggestion from me was to use a bit deeper depth, like 150 or maybe even 500 (or even safer 1000?). I suppose the depth number corresponds to the number of commits to go back since HEAD. So I guess on average we probably do around 100 - 150 commits/merges between deploys to live, but this number is volatile and can vary depending on different devs and how often they commit etc. So maybe going with a deep depth like 1000 is a good idea? Another project I'm currently working on has ~2300 commits, so setting to 1000 depth should more than halve the history.

bboro: We could also start squashing PR merge commits which should reduce the depth to just 1 commit per PR merge

amitaibu: Let's start with 500 and see where it gets us.

> We could also start squashing PR merge commits 

tbh, I've been doing it for a while - yeah, we could ask all others to do that as well

bboro: Is there a way to change the default merge button to be "Squash and Merge" without having to open the dropdown? :smile:

amitaibu: You need to do it only once per repo

amitaibu (approved):

amitaibu:
Thanks

PR #408, opened 2023-01-11
Added bottom padding to .slick-track.

balagan73:
I have added a bottom padding to the carousel, otherwise box-shadow might get truncated.


bboro:
Did you mean to close this?

amitaibu:
Yes, on Drupal-starter we don't have shadow-box on the card, so it's not needed

Issue #383, opened 2022-12-07
Memory limit also affects PHPunit tests


mariano-dagostino:
```
 ✔ Some tests result:

PHP Fatal error:  Allowed memory size of 268435456 bytes exhausted (tried to allocate 20480 bytes) in /var/www/html/vendor/symfony/dependency-injection/Definition.php on line 521
```

It seems we should run phpunit without memory limits. This is not replicable in drupal-starter, you need a project with more tests.

PR #409, opened 2023-01-13
Memory limit for CLI

AronNovak:
#383

### DIFF
--- a/.ddev/config.local.yaml.example
+++ b/.ddev/config.local.yaml.example
@@ -48,4 +48,5 @@ hooks:
     # Web requests should be still limited, as the
     # hosting environment won't provide unlimited
     # memory either.
+    # @see https://github.com/drud/ddev/issues/1825#issuecomment-529964728
     - exec: perl -pi -e 's/memory_limit.*$/memory_limit = -1/' /etc/php/${DDEV_PHP_VERSION}/cli/conf.d/*.ini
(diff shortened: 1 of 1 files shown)


AronNovak:
image

.ddev/config.local.yaml.example
```diff
@@ -42,3 +42,10 @@ hooks:
 
     # Login as admin.
     - exec-host: ddev login
+
+    # Make sure PHP CLI has no memory limit, this is
+    # especially for unit testing.
+    # Web requests should be still limited, as the
+    # hosting environment won't provide unlimited
+    # memory either.
+    - exec: perl -pi -e 's/memory_limit.*$/memory_limit = -1/' /etc/php/${DDEV_PHP_VERSION}/cli/conf.d/*.ini
```
amitaibu, line 51: Can we do something similar to this - https://stackoverflow.com/a/72787008/750039 ?

AronNovak: I'd not do that. It affects everything, web requests and CLI too. For generating web pages, we never want to allow that amount of memory to be consumed and I think it's better to enforce it early and not only on Pantheon.

See this:  https://github.com/drud/ddev/issues/1825#issuecomment-529964728 - it's @rfay approved solution :)

amitaibu: Got it. Let's add the link on top of the code then

amitaibu (approved):

PR #410, opened 2023-01-20
Fix js deletions on travis retry

bboro:
Replaced `patchwork/jsqueeze` with `coderaiser/minify`. As this is an npm package, and we automatically run npm install before each compile, this should fix JS files getting deleted due to composer dev dependencies being uninstalled (or cached as uninstalled) on travis retries.


amitaibu (approved):
:100:

PR #418, opened 2023-01-30
Preserve "Drupal" in js minify

bboro:
This is because Locale module looks for `Drupal.t` in order to decide whether to load the string translation for a phrase. (see `_locale_parse_js_file`)

If `Drupal.t('Read more')` is mangled to something else during minification, like `a.t('Read more')` then locale doesn't add the string translation anymore for that phrase.


AronNovak (approved):
:+1: makes a lot of sense

Issue #417, opened 2023-01-27
Repo cloner command [5h]


AronNovak:
We could improve the tooling for starting a new project.
In addition of what we have for the autodeploy, we could take care of:
 - site name replacement
 - configuring the proper docker mirror
 - preparing the Pantheon repo to be usable
 - initialize the QA env on Pantheon
 - install the Drupal site on QA
 - enable Redis
 - add site locking

The signature could be:

```
ddev robo project:clone [new GH repo URL] [pantheon project] [pantheon token] [github token] [docker mirror URL] [http basic auth user] [http basic auth password]
```

Then the manual steps would be:
 - create a new Pantheon project
 - add a new Pantheon user and generate the key

Should we invest in this?

amitaibu:
Yes, I like the idea. Let's talk about it before starting.

PR #422, opened 2023-02-02
Bootstrap project command - phase II

AronNovak:
#417

### DIFF
--- a/pantheon_template/settings.pantheon.php
+++ b/pantheon_template/settings.pantheon.php
@@ -1,0 +1,204 @@
+<?php
+
+/**
+ * @file
+ * Pantheon configuration file.
+ *
+ * IMPORTANT NOTE:
+ * Do not modify this file. This file is maintained by Pantheon.
+ *
+ * Site-specific modifications belong in settings.php, not this file. This file
+ * may change in future releases and modifications would cause conflicts when
+ * attempting to apply upstream updates.
+ */
+
+/**
+ * Version of Pantheon files.
+ *
+ * This is a monotonically-increasing sequence number that is
+ * incremented whenever a change is made to any Pantheon file.
+ * Not changed if Drupal core is updated without any change to
+ * any Pantheon file.
+ *
+ * The Pantheon version is included in the git tag only if a
+ * release is made that includes changes to Pantheon files, but
+ * not to any Drupal files.
+ */
+if (!defined("PANTHEON_VERSION")) {
+  define("PANTHEON_VERSION", "4");
+}
+
+/**
+ * Determine whether this is a preproduction or production environment, and
+ * then load the pantheon services.yml file.  This file should be named either
+ * 'pantheon-production-services.yml' (for 'live' or 'test' environments)
+ * 'pantheon-preproduction-services.yml' (for 'dev' or multidev environments).
+ */
+$pantheon_services_file = __DIR__ . '/services.pantheon.preproduction.yml';
+if (
+  isset($_ENV['PANTHEON_ENVIRONMENT']) &&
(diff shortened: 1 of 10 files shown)


AronNovak:
image

AronNovak:
image

AronNovak:
Now it works correctly to
 - prepare GH repo
 - create new Pantheon project
 - prepare Pantheon project
 - create QA env
 - install environments
 - lock environments

What still needs to be done:
 - integrate autodeploy configuration, so it does not have to be a different command invocation after project setup.

robo-components/BootstrapTrait.php
```diff
@@ -0,0 +1,311 @@
...
+    $this->verifyRequirements($project_name, $organization, $project_machine_name, $terminus_token, $github_token, $docker_mirror_url, $http_basic_auth_user, $http_basic_auth_password);
+
+    $this->prepareGithubRepository($project_name, $organization, $project_machine_name, $github_repository_url, $docker_mirror_url);
+
+    $this->createPantheonProject($terminus_token, $project_name, $project_machine_name, $organization);
+
+    $this->deployPantheonInstallEnv('dev', $project_machine_name);
+    $this->deployPantheonInstallEnv('qa', $project_machine_name);
+
+    if ($http_basic_auth_user && $http_basic_auth_password) {
```
amitaibu, line 46: Let's move this check into `lockPantheonEnvironments`

robo-components/BootstrapTrait.php
```diff
@@ -0,0 +1,311 @@
...
+      $this->lockPantheonEnvironments($project_machine_name, $http_basic_auth_user, $http_basic_auth_password);
+    }
+    else {
+      $this->say("No HTTP basic auth credentials were provided. Pantheon environments will not be locked.");
+    }
+
+    $this->say("Bootstrap completed successfully.");
+    $this->say("You might want to run the following commands to properly place the project:");
+    $this->say("mv .bootstrap ../$project_machine_name");
```
amitaibu, line 55: What's inside `.bootstrap`?

Also, maybe we should do it for them?

amitaibu (commented):
Great stuff, one comment

robo-components/BootstrapTrait.php
AronNovak: We are bound to what's mounted into DDEV, that's why we cannot easily move it, but i will think about it, also we have an issue related to invoke the autodeploy config, what would be executed in another DDEV project, that needs to be refactored to be able to do it in one step.

AronNovak: Inside it's the new project's GH repository.

pantheon_template/settings.pantheon.php
```diff
@@ -0,0 +1,204 @@
+<?php
```
mariano-dagostino, line 1: @AronNovak Should we download this file directly from https://github.com/pantheon-systems/drops-8/blob/495b94f873878ee0b06651307059eccab337562d/sites/default/settings.pantheon.php

mariano-dagostino (approved):

pantheon_template/settings.pantheon.php
AronNovak: @mariano-dagostino Good point. Do you have a suggestion on which branch to use? 
https://github.com/pantheon-systems/drops-8/tree/default-9.x/sites is not the same that you linked for instance, it seems to be a bit chaotic.

amitaibu: I don't think it will change much, so we can start with the file we have

amitaibu:
Thanks

PR #426, opened 2023-02-07
Translation management

bboro:


bboro:
Ported from another project.

Looks like whoever set up the languages originally (may be me) simply copied `config/sync/language/ar` into `config/sync/language/es`, so all Spanish config was coming up as Arabic. Had me wondering there for a bit. Fixed that now so there's 120 files diff due to that.

TB: 1.5h
TB:

amitaibu:
> Looks like whoever set up the languages originally

Oopsy, it was probably me :sweat_smile:

amitaibu:
Thanks

PR #434, opened 2023-02-16
Add `buildElementCarousel` and improve theme name

mariano-dagostino:
Screenshot-2023-02-16_16-11


amitaibu:
I've added `ElementTrait::buildElementCarousel` to keep it inline with other code: Style guide and PEVB should both call the same helper method, rather than calling the theme file directly. This, in turn, also gives us some type safety on top of the theme.

image

amitaibu:
@mariano-dagostino in your screenshot there's padding missing. Are you able to reproduce, or is it because `ddev theme:watch` was not running?

image

amitaibu (approved):

PR #442, opened 2023-03-01
Added alignment and credit field to MediaCaptionTrait

balagan73:
MediaCaptionTrait now can display the credit field and can handle alignment.
Credits to @dipakmdhrm


web/modules/custom/server_general/src/MediaCaptionTrait.php
```diff
@@ -21,19 +21,27 @@ trait MediaCaptionTrait {
...
+    $credit = $entity->hasField($credit_field_name) ? $this->getTextFieldValue($entity, $credit_field_name) : NULL;
+    if (empty($caption) && empty($credit)) {
+      // Nothing to output.
       return [];
     }
 
     return [
       '#theme' => 'server_theme_media_caption',
       '#caption' => $caption,
+      '#credit' => $credit,
```
amitaibu, line 43: I think we should change this to $items - that will get a render array. So we can pass caption, credit, etc'.

Then we could use the `verticalSpacing` -- we can pass alignment there.

amitaibu (changes requested):
Thanks. Screenshots would be nice :)

In any case, I think we should change strategy, and move to composing the elements together

web/modules/custom/server_general/src/MediaCaptionTrait.php
balagan73: image

verticalSpacingTiny has gap-y-2 class, and here we have 8px gap. Should we create another vertical spacer with gap-y-1 class?

amitaibu: Let's use `verticalSpacingTiny` as is

web/themes/custom/server_theme/server_theme.theme
```diff
@@ -399,6 +399,7 @@ function server_theme_theme() {
   $info['server_theme_media_caption'] = [
     'variables' => [
       'caption' => NULL,
+      'credit' => NULL,
```
amitaibu, line 402: => `$items => []`

balagan73:
Then we can probably get rid of the templates for the caption, video and media.

balagan73:
Style guide after the updates.

image

balagan73:
Maybe deleting the video and image templates is not a good idea. Added wrappers for caption font colors.

web/themes/custom/server_theme/server_theme.theme
amitaibu: ^^. We should remove caption and credit, and call it `$items`

web/modules/custom/server_style_guide/src/Controller/StyleGuideController.php
```diff
@@ -285,17 +285,27 @@ protected function getCardsCentered(): array {
...
+    $elements[] = $caption;
+
+    // Credit.
+    $credit = $this->wrapTextResponsiveFontSize('Photo Credits: John Doe', 'base');
+    $credit = $this->wrapTextColor($credit, 'dark-gray');
+    $elements[] = $this->wrapTextItalic($credit);
+
+    $items = $this->wrapContainerVerticalSpacingTiny($elements, 'start');
 
     return [
```
amitaibu, line 305: We shouldn't be calling these theme functions form StyleGuide - we should call the method on the trait. You can mock the $entity for that

balagan73:
I will start again in a new branch, this went in a wrong way.

PR #444, opened 2023-03-03
New max width wrapper + TW scan js files

andreytroeglazov:


web/themes/custom/server_theme/templates/server-theme-container-max-width.html.twig
```diff
@@ -0,0 +1,13 @@
...
+  {% set size_classes = 'max-w-lg' %}
+{% elseif width == 'xl' %}
+  {% set size_classes = 'max-w-xl' %}
+{% elseif width == '2xl' %}
+  {% set size_classes = 'max-w-2xl' %}
+{% elseif width == '3xl' %}
+  {% set size_classes = 'max-w-3xl' %}
+{% endif %}
+
+<div class="{{ size_classes }} mx-auto">
```
amitaibu, line 11: I could imagine a different middleCenter that wold add the mx-auto, so we can remove from here

```suggestion
<div class="{{ size_classes }}">
```

web/modules/custom/server_general/src/ElementWrapTrait.php
```diff
@@ -462,6 +462,31 @@ protected function wrapTextColor(array|string|TranslatableMarkup $element, strin
...
+   *
+   * @param array|string|\Drupal\Core\StringTranslation\TranslatableMarkup $element
+   *   The render array, string or a TranslatableMarkup object.
+   * @param string $width
+   *   Max width. Allowed values are `lg`, `xl`, `2xl` and `3xl.
+   *
+   * @return array
+   *   Render array.
+   */
+  protected function wrapContainerMaxWidth(array|string|TranslatableMarkup $element, string $width): array {
```
amitaibu, line 476: Place above next to add `wrapContainer...`

web/themes/custom/server_theme/templates/server-theme-container-max-width.html.twig
andreytroeglazov: Done

amitaibu (approved):
Thanks, MWG

PR #453, opened 2023-03-13
Paragraph translation sanity check

AronNovak:


web/modules/custom/server_general/tests/src/ExistingSite/ServerGeneralParagraphTranslatability.php
```diff
@@ -0,0 +1,34 @@
...
+
+use weitzman\DrupalTestTraits\ExistingSiteBase;
+
+/**
+ * Tests entity references.
+ */
+class ServerGeneralParagraphTranslatability extends ExistingSiteBase {
+
+  /**
+   * Checks all entity reference fields to nodes use the select2 widget.
```
AronNovak, line 13:
```suggestion
   * Checks all entity reference revision fields for Paragraphs are non-translatable.
```

web/modules/custom/server_general/tests/src/ExistingSite/ServerGeneralParagraphTranslatability.php
```diff
@@ -0,0 +1,34 @@
...
+
+/**
+ * Tests entity references.
+ */
+class ServerGeneralParagraphTranslatability extends ExistingSiteBase {
+
+  /**
+   * Checks all entity reference revision fields for Paragraphs are non-translatable.
+   */
+  public function testParagraphsTranslation() {
```
amitaibu, line 15: Another idea that might be better - can we throw an Exception if one tries to set the field incorrectly?

amitaibu: Good point :+1: Save ourselves from ourselves :)

web/modules/custom/server_general/tests/src/ExistingSite/ServerGeneralParagraphTranslatability.php
```diff
@@ -0,0 +1,34 @@
...
+        continue;
+      }
+
+      $storage_definition = $field_config->getFieldStorageDefinition();
+      $referencing = $storage_definition->getSetting('target_type');
+      if ($referencing != 'paragraph') {
+        continue;
+      }
+
+      $this->assertFalse($field_config->isTranslatable(), sprintf("%s must not be translatable", $id));
```
amitaibu, line 30:
```suggestion
      $this->assertFalse($field_config->isTranslatable(), sprintf("%s must not be translatable as it is a Paragraph field", $id));
```

amitaibu (approved):

PR #464, opened 2023-03-21
Modernize PEVB plugins

amitaibu:
- All PEVBs are now calling `buildElement...`, so they are all having consistent code
- Added new elements to Style guide
- `\Drupal\server_style_guide\Controller\StyleGuideController::buildProcessedText` now wraps text with `prose`, like we do on PEVB.

| Element | Style guide | PEVB |
| -- | -- | -- |
| Search | image | image |
| Carousel | image | image | 
| Paragraph title and content (Used by `Text` paragraph type) | image | image |

### DIFF
--- a/web/modules/custom/server_general/src/Plugin/EntityViewBuilder/ParagraphSearch.php
+++ b/web/modules/custom/server_general/src/Plugin/EntityViewBuilder/ParagraphSearch.php
@@ -7,8 +7,7 @@ namespace Drupal\server_general\Plugin\EntityViewBuilder;
 use Drupal\Core\Block\BlockManagerInterface;
 use Drupal\paragraphs\ParagraphInterface;
 use Drupal\pluggable_entity_view_builder\EntityViewBuilderPluginAbstract;
-use Drupal\server_general\ButtonTrait;
-use Drupal\server_general\ElementWrapTrait;
+use Drupal\server_general\ElementTrait;
 use Drupal\server_general\EmbedBlockTrait;
 use Symfony\Component\DependencyInjection\ContainerInterface;
 use Symfony\Component\HttpFoundation\Request;
@@ -24,9 +23,8 @@ use Symfony\Component\HttpFoundation\Request;
  */
 class ParagraphSearch extends EntityViewBuilderPluginAbstract {
 
+  use ElementTrait;
   use EmbedBlockTrait;
-  use ButtonTrait;
-  use ElementWrapTrait;
 
   /**
    * The machine name of the facets to show.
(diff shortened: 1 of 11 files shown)


amitaibu:
Much nicer now - PEVB and Style guide pass raw data, and ElementTrait is responsible for the theming. If you are coming from Elm think of Model (holding data0 & View (rendering).

web/modules/custom/server_general/src/Plugin/EntityViewBuilder/ParagraphSearch.php
```diff
@@ -81,17 +79,20 @@ public static function create(ContainerInterface $container, array $configuratio
    *   Render array.
    */
   public function buildFull(array $build, ParagraphInterface $entity): array {
-    $elements = [];
-
-    // Search term and facets.
-    $elements[] = $this->buildSearchTermAndFacets();
+    // Facets.
+    $facets_items = [];
+    foreach ($this->facetNames as $facet_name) {
```
mariano-dagostino, line 84: Where is this defined?

mariano-dagostino: I see we are just moving code around. just curious.

web/modules/custom/server_general/src/TitleAndLabelsTrait.php
```diff
@@ -2,38 +2,11 @@
 
 namespace Drupal\server_general;
 
-use Drupal\node\NodeInterface;
-use Drupal\pluggable_entity_view_builder\BuildFieldTrait;
-
 /**
  * Helper method for building Title and labels of a content.
  */
 trait TitleAndLabelsTrait {
```
mariano-dagostino, line 8: This class calls `wrapHtmlTag` should we add the ElementWrapTrait?

mariano-dagostino (changes requested):
@amitaibu Looks good. just one question on `TitleAndLabelsTrait`

web/modules/custom/server_general/src/Plugin/EntityViewBuilder/ParagraphSearch.php
amitaibu: https://github.com/Gizra/drupal-starter/blob/a83048fa3c2355a3908186f977374269003afd05/web/modules/custom/server_general/src/Plugin/EntityViewBuilder/ParagraphSearch.php#L31-L45

web/modules/custom/server_general/src/TitleAndLabelsTrait.php
amitaibu: It seems to be working without (I believe because trait is in the same namespace), but added to be explicit.

Issue #447, opened 2023-03-07
Config not clean after clean installation [2h]


amitaibu:
The Arabic/ Spanish issue is still around.

![image](https://user-images.githubusercontent.com/125707/223414303-8ddc60ef-1742-4387-9622-3f3eb6c61a1f.png)

GizraNotifier:
The latest merged PR #471 just got deployed successfully to Pantheon [`qa`](https://gizra-drupal-starter-qa.pantheonsite.io) environment

PR #471, opened 2023-03-24
Delete all config translations. And do not pull translations from Drupal on install

mariano-dagostino:
#447


amitaibu:
Thanks. I'm not sure about this approach - as it's not fixing the root cause, it's only delaying it for implementing projects that would need a multilingual site.

Maybe we could keep the part that skips the import - it makes `ddev restart` faster. But we should keep the language translations, so upon `ddev restart` one we switch to Arabic or Spanish, we'd see the site translated.

mariano-dagostino:
@amitaibu I also tried https://www.drupal.org/files/issues/2023-01-31/2806009-116.patch without success...

it seems some sort of race condition.

1.5h/2 from my side

amitaibu:
@AronNovak care to give it a go? Bump to total 4h

AronNovak:
@amitaibu Sure, especially that https://www.drupal.org/files/issues/2023-01-31/2806009-116.patch was committed to 9.5.x a few days ago

AronNovak:
@amitaibu I finish https://github.com/Gizra/drupal-starter/pull/485 first to have the latest updates in, then I can continue here.

AronNovak:
@amitaibu I think this PR might be good as-is.
For out-of-the-box translations, we have:
https://github.com/Gizra/drupal-starter/tree/main/config/po_files

and a Robo command to import those.
https://github.com/Gizra/drupal-starter/blob/main/.ddev/config.local.yaml.example#L19

I'll make sure those existing PO files are wired properly and has meaningful translations inside.

PR #478, opened 2023-03-27
Add Two-factor Authentication (TFA) to `Administrator` and `Content editor` users

amitaibu:
Making login for privileged users more secure, by using "Google Authenticator" app (or similar). We could still log into those users with for example `ddev drush uli --uid=4`

On the first login of the user they are prompted to add the authentication. If they don't do that, on their next login they will be blocked, and admin would need to re-grant them access.

| Item |
| -- |
|  image |
| image | 
| image | 
| image | 
| image |
| image |


.ddev/config.local.yaml.example
```diff
@@ -9,6 +9,9 @@ hooks:
     # Private files directory.
     - exec: mkdir /var/www/private
 
+    # Create a private key for Two-factor Authentication (TFA).
+    - exec: openssl rand -base64 32 > /var/www/tfa.key
```
amitaibu, line 13: @AronNovak keep here, or move to [web-build/Dockerfile](https://github.com/Gizra/drupal-starter/blob/6023ee556f51676bbd87cb76d4056e84b50d9c62/.ddev/web-build/Dockerfile)?

README.md
```diff
@@ -388,3 +388,15 @@ To import the config translations:
...
+## Two-factor Authentication (TFA)
+
+TFA is enabled for the Administrator and Content editor users.
+The default settings under `/admin/config/people/tfa` define "Skip Validation" is 1. That is,
+when a privileged user will login, they must enable their TFA. Otherwise, on a second
+login, they will already be blocked. A site admin may reset their validation tries
+under the `/admin/people` page.
+The TFA method that is enabled is one that uses Google authenticator (or similar).
+
+The TFA private key is created under `/var/www/tfa.key`.
```
amitaibu, line 401: @AronNovak How should we handle this on Pantheon? There's also the option of passing it as ENV variable if easier

web/themes/custom/server_theme/templates/status-messages.html.twig
```diff
@@ -30,7 +30,7 @@
   {% elseif type == 'status' %}
     {% set message_class = 'bg-green-400' %}
   {% endif %}
-  <div role="contentinfo" aria-label="{{ status_headings[type] }}"{{ attributes|without('role', 'aria-label') }} class="{{ message_class }} messages--{{ type }} break-all text-white py-2 px-4 rounded rounded-large text-left mt-4">
+  <div role="contentinfo" aria-label="{{ status_headings[type] }}"{{ attributes|without('role', 'aria-label') }} class="{{ message_class }} messages--{{ type }} text-white py-2 px-4 rounded rounded-large text-left mt-4 [&_a]:underline">
```
amitaibu, line 33: We make sure the `a href` are visible on the HTML

.ddev/config.local.yaml.example
amitaibu: Ignore, changed to be an ENV variable

README.md
amitaibu: Ignore, changed to be an ENV variable

web/themes/custom/server_theme/src/pcss/form.pcss
```diff
@@ -0,0 +1,15 @@
+form {
```
amitaibu, line 1: image

README.md
AronNovak: There's a trade-off between security and DX.
https://docs.pantheon.io/guides/environment-configuration/read-environment-config - Pantheon suggests that we should put secrets in a JSON in the private filesystem. It makes sense, but it's harder to manage.
Also it is possible that we follow the current path and we put it in `settings.php`, what gets synced between GH and Pantheon.

A rigorous pentest would not be fond of storing any kind of secret in a GH repository. At the very first ElasticSearch project, we were asked to store the ES secrets in the private filesystem, then we did optimize it for the sake of DX.
It would be nicer that all the secrets would be guarded by Pantheon, so a full-blown access to Pantheon would be needed to read all these sensitive data.

amitaibu: There seems to be a newer plugin - https://github.com/pantheon-systems/terminus-secrets-manager-plugin

AronNovak: Yep, that plugin just assembles that JSON file essentially, we can indeed use it. Last time somehow I could not make it working, but it's certainly doable.

AronNovak (approved):

amitaibu:
Reverted to use a file.

@AronNovak handing PR to you to wire the Pantheon deploy/ site create part to work with a private file with the TFA .

AronNovak:
@amitaibu I will add the infra part today or tomorrow.

AronNovak:
https://docs.pantheon.io/guides/integrated-composer
It seems for the Secrets Manager, we must use Integrated Composer workflow, which is an out of scope conversion at the moment.

amitaibu:
So we'll use https://github.com/pantheon-systems/terminus-secrets-plugin instead?

AronNovak:
@amitaibu Yep, I just tried that out, it works well.

amitaibu:
Even better, so anyone could fork the starter and won't be blocked by getting early access

AronNovak:
@amitaibu It seems Pantheon part is ready. We use the environment-variable way, at DDEV, we read it from the raw file, on Pantheon, we read it from the JSON that the tool provides us.

amitaibu:
> at DDEV, we read it from the raw file

Let's use this then, instead of reading from the file

https://github.com/Gizra/drupal-starter/pull/478/commits/84a1df8f61a4e6a300f14dcdc388cad75e7cc2c2

AronNovak:
@amitaibu I utilized DDEV global config to solve it coherently for local dev environments, and still, we don't hard-wire it in the GH repo.

amitaibu:
Thanks :)

PR #485, opened 2023-03-31
PHP 8.1 + Mass Composer package updates

AronNovak:


AronNovak:
We should use PHP 8.1-compatible things only (soon).

https://www.php.net/supported-versions.php

amitaibu (approved):

amitaibu:
Thanks

PR #488, opened 2023-04-13
Text styles (prose and non-prose) on Style guide page

amitaibu:
| Non prose | Prose |
| -- | -- |
| image | image |


web/modules/custom/server_style_guide/src/Controller/StyleGuideController.php
```diff
@@ -537,6 +539,25 @@ protected function getTextDecorations(): array {
...
+  /**
+   * Get text styles prose, and non-prose.
+   *
+   * @return array
+   *   A render array.
+   */
+  protected function getTextStyles(): array {
+    $build = [];
+
+    // Font weight for a string.
```
amitaibu, line 551:
```suggestion
```

mariano-dagostino (changes requested):
Do we want to standardize the sizes of headers for both options?

Screenshot_2023-04-13_15-35-10

amitaibu:
@isapisa @nora-novak What do you think about Mariano's comment - should the H1-H5 be the same in prose and non prose?

amitaibu:
We are now showing the same font size for headings (h1- h5) on prose and non-prose, as we wrap all of them with `wrapProseText`.

The non-prose version should not have a margin applied to it - and that's the case.

image

mariano-dagostino (approved):

web/modules/custom/server_style_guide/templates/server-style-guide-text-styles.html.twig
```diff
@@ -0,0 +1,61 @@
+<h1>This is an example for H1</h1>
+<h2>This is an example for H2</h2>
+<h3>This is an example for H3</h3>
+<h4>This is an example for H4</h4>
+<h5>This is an example for H5</h5>
```
amitaibu, line 5:
```suggestion
<h1>This is an example for h1</h1>
<h2>This is an example for h2</h2>
<h3>This is an example for h3</h3>
<h4>This is an example for h4</h4>
<h5>This is an example for h5</h5>
```

Issue #441, opened 2023-02-28
Provide a vanila js slider instead of slick


mariano-dagostino:
As drupal evolves to remove the jquery dependency as much as possible I think is worth to analyze if we can remove jQuery from the carousels

https://github.com/davidhund/awesome-vanilla-js#carousel

amitaibu:
About 6 months ago I tried to look for a replacement to Slick, but none was good enough. Yeah, I won't mind changing it

amitaibu:
Tried Swiper JS, but I don't think its worth the effort - https://github.com/Gizra/drupal-starter/pull/491#issuecomment-1532781449

bmunslow:
As an idea, Splide is looking good (though I haven't tried it myself):

- https://github.com/Splidejs/splide
- https://www.drupal.org/project/splide

sonvir249:
@amitaibu @bmunslow I have an Idea of replacing slick slider with [tiny slider](https://github.com/ganlanyuan/tiny-slider) or [glider-js](https://glidejs.com/docs/). Please do let me know your thoughts. Thanks!

amitaibu:
Unless there's a clear advantage, I believe we can keep the current implementation.

sonvir249:
@amitaibu As Drupal is removing the jquery dependency we can also do the same as mentioned by @mariano-dagostino for slider. Also I can see jQuery.once() is used in custom `server_theme`. jQuery.once() with javascript once().

amitaibu:
Sure, PRs are welcome 😄

PR #491, opened 2023-04-13
Replace slick with Swiper js

amitaibu:
#441


amitaibu:
Early work

image

amitaibu:
Closing as I'm not convinced that it's worth the effort. Swiper has its own peculiarities, and I don't believe it's a significantly superior solution.

Issue #280, opened 2022-08-11
Add seckit and HSTS [3h]


amitaibu:
@mariano-dagostino I believe you've implemented it on other projects. Do you see any downside in including it by default with Drupal-stater?

mariano-dagostino:
@amitaibu The only issue I see is you need to be aware of seckit blocking remote URL calls by your js code. Therefore if you have some elm or js app that do a get requests to an external api, you need to whitelist it in the seckit config.

This also applies for iframe content, so google maps, youtube embeds, etc need to be whitelisted as well.

amitaibu:
> you need to whitelist it in the seckit config.

Let's add that to the README.

> so google maps, youtube embeds, etc

Good point. Let's add to the README, but also let's whitelist those services (Youtube, Maps, Google Analytics, come to mind)

mariano-dagostino:
@amitaibu I've been checking this today, it seems there is a big catch. If you use Ckeditor you need to enabled script-src: 'unsafe-inline` which is not a good practice.

I've been doing some research and there is an alternative module to seckit called [CSP](https://www.drupal.org/project/csp/) I've never used it, but it seems it adds this unsafe-inline option conditionally when CKeditor is in place.

Anyways now rethinking my [answer](https://github.com/Gizra/drupal-starter/issues/280#issuecomment-1211892892) I think Seckit is not a good candidate for drupal starter. So either we try CSP, or close this issue.

1h

amitaibu:
Seckit seems to be more popular and has more options to harden the site. Can you give it ~2h to see what would be a possible path for it and CKeditor.

mariano-dagostino:
@amitaibu We may have more luck with CKEditor 5 (now included in core) https://ckeditor.com/docs/ckeditor5/latest/installation/advanced/csp.html 

![Screenshot_2022-10-24_10-49-04](https://user-images.githubusercontent.com/5313631/197541675-48d2804a-9257-4cbb-988a-8e581b48c215.png)

I don't know if CKEditor for core is already doing this. I checked the issue queue and haven't found anything related to this topic. Anyways seems the most promising path of action.

- Check if CKEditor 5 is already using external CSS files.
- Configure Seckit accordingly.
- If CKeditor is not using external CSS files, open a drupal issue to request this change.

+0.5h

amitaibu:
@mariano-dagostino let's revive this when you have time.

mariano-dagostino:
[...]

PR #492, opened 2023-04-14
Add anchor links to all headings

amitaibu:
image

### DIFF
--- a/web/themes/custom/server_theme/server_theme.libraries.yml
+++ b/web/themes/custom/server_theme/server_theme.libraries.yml
@@ -16,6 +16,7 @@ global-scripts:
     - core/jquery
     - core/drupal
     - core/jquery.once
+    - server_theme/anchor
 
 
 # "View more" button.
@@ -68,3 +69,12 @@ slick:
       /libraries/slick-carousel/slick/slick-theme.css: { weight: -9 }
   dependencies:
     - core/jquery
+
+# Add anchors to headings.
+anchor:
+  js:
+    /libraries/anchor-js/anchor.js: {}
+  dependencies:
+    - core/jquery
+    - core/drupal
+    - core/jquery.once
--- a/web/themes/custom/server_theme/server_theme.libraries.yml
+++ b/web/themes/custom/server_theme/server_theme.libraries.yml
@@ -75,6 +75,4 @@ anchor:
   js:
     /libraries/anchor-js/anchor.js: {}
   dependencies:
-    - core/jquery
     - core/drupal
-    - core/jquery.once
--- a/web/themes/custom/server_theme/server_theme.libraries.yml
+++ b/web/themes/custom/server_theme/server_theme.libraries.yml
@@ -73,6 +73,6 @@ slick:
 # Add anchors to headings.
 anchor:
   js:
-    /libraries/anchor-js/anchor.js: {}
+    /libraries/anchor-js/anchor.min.js: {}
   dependencies:
     - core/drupal
--- a/web/themes/custom/server_theme/src/js/scripts.js
+++ b/web/themes/custom/server_theme/src/js/scripts.js
@@ -13,6 +13,10 @@
     attach: function (context, settings) {
       // Add anchor links to all headings.
       const anchors = new AnchorJS();
+      anchors.options = {
+        icon: '',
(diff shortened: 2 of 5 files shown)


amitaibu:
@isapisa @nora-novak What do you think?

isapisa:
@amitaibu why would we want to do this?

amitaibu:
This allows adding anchors to any h2 - h6 heading. Useful on long pages/ posts. Thanks to this we can for example link to this section https://www.gizra.com/jobs/candidates-info/#timebox

mariano-dagostino (changes requested):
@amitaibu Do we want to wrap this in an 'once()' call?

web/themes/custom/server_theme/server_theme.libraries.yml
```diff
@@ -68,3 +69,12 @@ slick:
...
     - core/jquery
+
+# Add anchors to headings.
+anchor:
+  js:
+    /libraries/anchor-js/anchor.js: {}
+  dependencies:
+    - core/jquery
+    - core/drupal
+    - core/jquery.once
```
amitaibu, line 80: I'm not sure if this is needed, I'll check

web/themes/custom/server_theme/src/js/scripts.js
```diff
@@ -10,6 +11,9 @@
    */
   Drupal.behaviors.themeServerGeneral = {
     attach: function (context, settings) {
+      // Add anchor links to all headings.
```
amitaibu, line 14: @mariano-dagostino Good call about `once`. I think it might not be needed since AnchorJs already does it [here](https://github.com/bryanbraun/anchorjs/blob/7820d5f6a6d41eb92000ab9c22b9362262a6b8a8/anchor.js#L260-L265)

web/themes/custom/server_theme/server_theme.libraries.yml
```diff
@@ -68,3 +69,10 @@ slick:
       /libraries/slick-carousel/slick/slick-theme.css: { weight: -9 }
   dependencies:
     - core/jquery
+
+# Add anchors to headings.
+anchor:
+  js:
+    /libraries/anchor-js/anchor.js: {}
+  dependencies:
+    - core/drupal
```
mariano-dagostino, line 78: Is this even required? The anchor.js seems a standalone script.

web/themes/custom/server_theme/server_theme.libraries.yml
```diff
@@ -68,3 +69,10 @@ slick:
       /libraries/slick-carousel/slick/slick-theme.css: { weight: -9 }
   dependencies:
     - core/jquery
+
+# Add anchors to headings.
+anchor:
+  js:
+    /libraries/anchor-js/anchor.js: {}
```
mariano-dagostino, line 76: This is not a minimized version. I wonder if makes sense to move to the theme and point to `/dist` instead.

mariano-dagostino (changes requested):
@amitaibu Finished my review.

web/themes/custom/server_theme/server_theme.libraries.yml
amitaibu: Done, changed to the minified file

amitaibu:
Re-rolled PR

mariano-dagostino (commented):
@amitaibu This looks good and the functionality seems to work ok locally. However https://www.bryanbraun.com/anchorjs/#dont-run-it-too-late mentioned something interesting, maybe the link is not there yet in a heavy based JS site. I think we need to play with the JS weight to load it as early as possible.

amitaibu:
> maybe the link is not there yet in a heavy based JS site

As I can't see the problem, let's get it in, and deal once we have a real use case.

mariano-dagostino:
@amitaibu I removed the custom font call as it will interfere with #280 also, there is one character that breaks the ids generation. 6721707

Finally some examples of how to use tailwind classes.

https://user-images.githubusercontent.com/5313631/236534218-4e8965c0-134f-4fad-a7fc-c68507292b57.mp4

mariano-dagostino (approved):

amitaibu:
Made the icon slightly bigger

image

Issue #483, opened 2023-03-29
Refactor deploy notification


AronNovak:
The DDEV ENV variable passing is not reliable, it breaks for more complex commit messages.
We might pass the commit message via a `gitignore`-ed text file to avoid deployment instability due to commit message - YAML syntax issues.

amitaibu:
Can you give an example of a problematic commit message?

AronNovak:
```
Global configuration:

instrumentation-opt-in=false

omit-containers=[dba]

web-environment=[ remove not:prose as it is broken (#214)

 Add header text styles, remove prose as it is broken

 More adjustments. Styleguide helper

 Fix coding standards,TERMINUS_MACHINE_TOKEN=[secure],TRAVIS_COMMIT_MESSAGE=Add header text styles]

mutagen-enabled=false
```

An example Travis output.

AronNovak:
https://github.com/ddev/ddev/issues/4791

GizraNotifier:
The latest merged PR #493 just got deployed successfully to Pantheon [`qa`](https://gizra-drupal-starter-qa.pantheonsite.io) environment

PR #493, opened 2023-04-14
Do not allow comma to break deploy notification

AronNovak:
#483 

`ddev config global --web-environment-add` treats comma as a separator, so it breaks the detection or the YAML syntax in some cases.

A few attempts:
```
ddev config global --web-environment-add="MOO=abc,def"
```

```
ddev config global --web-environment-add="MOO=abc\,def"
```

```
ddev config global --web-environment-add="MOOA=\"abc,def\""
```

None of these actually include the comma in the content of the environment variable.


amitaibu (approved):
Thanks

PR #494, opened 2023-04-14
Do not deploy txt files to Pantheon at all

AronNovak:
We'd leak lots of information in those files.
Version numbers, patches, and so on.


pantheon_template/gitignore-template
```diff
@@ -60,3 +60,4 @@ Thumbs.db
 # Things in the core directory that Drupal 8 commits in the repository.
 !core/**/*.gz
 
+*.txt
```
AronNovak, line 63:
```suggestion
*.txt
*.md
```

amitaibu (approved):

PR #508, opened 2023-05-08
Simplify phpcs code - eliminate redundancy

AronNovak:


AronNovak:
We'd have a bit less precise reporting, but this PR eliminates a risk that `ddev phpcs` is green and Travis is red or the opposite, which happened recently.

amitaibu:
> We'd have a bit less precise reporting

Can you please show a screenshot of what it looks like?

I'm not entirely clear on what the PR does. What problem does it fix, and how?

AronNovak:
@amitaibu 
image

That would be the output.

The problem:
https://github.com/Gizra/drupal-starter/blob/main/ci-scripts/test_coder.sh#L72
https://github.com/Gizra/drupal-starter/blob/main/robo-components/PhpcsTrait.php#L42

We maintain two sets of ignores, one in Bash, one in Robo.
If we forget to update both in sync, we have either green Travis when `ddev phpcs` shows that there are issues, or the opposite.
We solve it by only keeping the Robo variant, eliminating the shell script that implements exactly the same logic.

amitaibu (approved):
:+1:

PR #509, opened 2023-05-12
Robo traits - Make the structure coherent

AronNovak:
We had some robo traits under `translation_management`, and some others under `robo-components`. Now it's coherent


AronNovak:
^^ @amitaibu What do you think?

PR #512, opened 2023-05-26
Enable Seckit: Replace Ckeditor with Ckeditor 5 to have an usable wysiwyg editor and CSP

mariano-dagostino:
#280


config/sync/language/es/block.block.claro_secondary_local_tasks.yml
```diff
@@ -1,2 +1,2 @@
 settings:
-  label: 'Solapas secundarias'
+  label: 'Pestañas secundarias'
```
amitaibu, line 2: @mariano-dagostino Pestañas? Why, I don't get the metaphor. :smile:

amitaibu:
@AronNovak @bboro, any objection to getting this in?

bboro:
@amitaibu no objections

composer.json
```diff
@@ -57,6 +57,7 @@
         "drupal/redirect": "^1.6",
         "drupal/redis": "^1.6",
         "drupal/search_api": "^1.18",
+        "drupal/seckit": "^2.0",
```
AronNovak, line 60: Hm, it seems to be a great idea to add SecKit, but it's out of scope here, not? Also should we provide sensible default config? Shall we enable CSP by default?

mariano-dagostino: @AronNovak This is issue is about enabling Seckit. The problem is seckit does not work with CKEditor 4 because it requires `unsafe-inline` See https://github.com/Gizra/drupal-starter/issues/280#issuecomment-1289076788

CKEditor 5 does not requires this setting, this is why we are upgrading.

config/sync/seckit.settings.yml
```diff
@@ -37,9 +37,9 @@ seckit_clickjacking:
   x_frame_allow_from: ''
 seckit_ssl:
   hsts: true
-  hsts_subdomains: false
-  hsts_max_age: 1000
-  hsts_preload: false
+  hsts_subdomains: true
```
AronNovak, line 40: This would hurt us for ElasticBeanstalk-based projects a bit, but generally it's a good default, I agree.

AronNovak (approved):

config/sync/seckit.settings.yml
amitaibu: Can you please elaborate some more - what it does, and how it would have an impact on ES.

AronNovak: "It allows [web servers](https://en.wikipedia.org/wiki/Web_server) to declare that [web browsers](https://en.wikipedia.org/wiki/HTTP_cookie) (or other complying [user agents](https://en.wikipedia.org/wiki/User_agent)) should automatically interact with it using only [HTTPS](https://en.wikipedia.org/wiki/HTTPS) connections,"

By default, on EB we do not have HTTPS, so in some scenarios, it might prevent easily access the site on port 80 for debugging purposes.

amitaibu:
Thanks y'all

Issue #514, opened 2023-05-30
Use `default_content` module for installation content [4h]


bboro:
Use [`default_content`](https://www.drupal.org/project/default_content) for easier management of content on-install.

- Install and configure `default_content`
- Move `migrate` default content to `default_content` module's yml exports
  - I'm thinking it's best to move it to a dedicated module `server_default_content`, so we can disable/enable without having to worry about dependencies
- Update documentations on workflow on how to export/update/import the content

GizraNotifier:
The latest merged PR just got deployed successfully to Pantheon [`qa`](https://gizra-drupal-starter-qa.pantheonsite.io) environment

bboro:
@amitaibu closing issue as this is now done.

PR #516, opened 2023-06-07
Use default content to manage installation content

bboro:
#514 

TB: 1.4 / 4h


bboro:
A thought crossed my mind to delete the default content from `server_migrate`, however I think it's quite useful to leave that as an example for an actual migration project. We simply don't enable `server_migrate` now on install.

Please have a go locally to test, make sure to re-generate you local `.ddev/config.local.yaml` from the example file.

AronNovak (approved):
This looks great. Later on we should add Spanish and Arabic default content too.

PR #517, opened 2023-06-09
Check requirements upon deploy

AronNovak:


web/.htaccess
```diff
@@ -158,12 +158,12 @@ AddEncoding gzip svgz
     # Serve gzip compressed CSS files if they exist and the client accepts gzip.
     RewriteCond %{HTTP:Accept-encoding} gzip
     RewriteCond %{REQUEST_FILENAME}\.gz -s
-    RewriteRule ^(.*css_[a-zA-Z0-9-_])\.css$ $1\.css\.gz [QSA]
+    RewriteRule ^(.*css_[a-zA-Z0-9-_]+)\.css$ $1\.css\.gz [QSA]
```
amitaibu, line 161: Related to Drupal core upgrade?

amitaibu (approved):
Thanks. Screenshot and MWG

web/.htaccess
AronNovak: It is.

AronNovak:
image

amitaibu:
What is the Unsupported release? Meaning, modules are not up to date? If so, I'd relax this, as sometimes we'd do deploys before upgrading modules

AronNovak:
@amitaibu Nope. Unsupported release is not oudated release. It means that the maintainer marked explicitly as Unsupported. Only 2 packages were like that. Simple outdated package does not throw a requirement error.

PR #522, opened 2023-06-30
Merge linting jobs to eliminate redundant `composer install`

AronNovak:


AronNovak:
As we have `ddev phpcs` and `ddev phpstan`, that works in the very same way, fine-grained reporting is not that crucial in my opinion.

AronNovak:
@amitaibu Worth it?

amitaibu:
I think it's nicer to keep separated, so just by looking at the test name I know which linter is broken. So IMO won't fix.

PR #524, opened 2023-07-03
Lint settings.php for Pantheon

AronNovak:
We had an incident on a client project, where it would have helped to get it through phpstan.


AronNovak:
^^ @amitaibu What do you think?

PR #531, opened 2023-07-30
Add deprecation error rule to PHPStan

dipakmdhrm:
While working on D10 upgrade issues I realize we should have PHPStan warn us of use of deprecated code as we work instead of having to remove them when we do upgrades.


bboro (approved):

bboro:
Nice!

PR #532, opened 2023-07-31
Use tearDown to unset variables to release memory during tests

bboro:


web/modules/custom/server_general/tests/src/ExistingSite/ServerGeneralTestBase.php
```diff
@@ -0,0 +1,35 @@
...
+
+/**
+ * Custom base class tailored for the site specifics.
+ *
+ * All non-js tests should extend this class instead of ExistingSiteBase.
+ */
+abstract class ServerGeneralTestBase extends ExistingSiteBase {
+
+  /**
+   * Tear down and unset variables.
```
amitaibu, line 19: Let's add a comment on why we have this. Possibly also a link to where it was copy/ pasted from for more context.

amitaibu (approved):
👌

PR #543, opened 2023-08-10
Fix project name replacement while bootstrap a new project

AronNovak:


robo-components/BootstrapTrait.php
```diff
@@ -129,7 +129,7 @@ protected function prepareGithubRepository(string $project_name, string $organiz
       ->run();
 
     $this->taskReplaceInFile('.bootstrap/.ddev/providers/pantheon.yaml')
-      ->from('yourproject.dev')
+      ->from('gizra-drupal-starter.qa')
```
AronNovak, line 132: See https://github.com/Gizra/drupal-starter/blob/main/.ddev/providers/pantheon.yaml#L40

andreytroeglazov (approved):

PR #548, opened 2023-08-22
Make needs users overridable for elasticsearch provisioning

AronNovak:


AronNovak:
^^ @mgurjanov That would be the way for more comfortable multidev ES provisioning.

robo-components/ElasticSearchTrait.php
```diff
@@ -77,10 +77,12 @@ protected function randomStr(
    *   The password of the ES admin user.
    * @param string|null $environment
    *   The environment ID. To test changes in the index config selectively.
+   * @param bool|null $needs_users_override
+   *   If set to 1 or TRUE, we do create a secrets.json file.
```
amitaibu, line 81:
```suggestion
   *   Set to TRUE to create a secrets.json file.
```

robo-components/ElasticSearchTrait.php
```diff
@@ -77,10 +77,12 @@ protected function randomStr(
...
    *
    * @throws \Exception
    */
-  public function elasticsearchProvision(string $es_url, string $username, string $password, ?string $environment = NULL): void {
+  public function elasticsearchProvision(string $es_url, string $username, string $password, ?string $environment = NULL, ?bool $needs_users_override = NULL): void {
```
amitaibu, line 85:
```suggestion
  public function elasticsearchProvision(string $es_url, string $username, string $password, ?string $environment = NULL, ?bool $needs_users_override = FALSE): void {
```

robo-components/ElasticSearchTrait.php
AronNovak: Right, but it will turn into a Bash argument, where writing TRUE is strange, it will be casted to PHP bool anyways, so TRUE is just a "TRUE" string, which is indeed casted to TRUE.

amitaibu: Yes, I assumed that was the cause, but I think we don't document it in other places that it's `1`. From the PHP perspective it's a `bool` value and should be TRUE/ FALSE.

robo-components/ElasticSearchTrait.php
```diff
@@ -77,10 +77,12 @@ protected function randomStr(
    *   The password of the ES admin user.
    * @param string|null $environment
    *   The environment ID. To test changes in the index config selectively.
+   * @param bool|null $needs_users_override
```
amitaibu, line 80:
```suggestion
   * @param bool $needs_users_override
```

amitaibu (approved):
Thanks

Issue #549, opened 2023-08-22
Javascript errors since Drupal 10 upgrade


bmunslow:
Multiple Javascript errors are thrown in a clean drupal-starter install with Drupal 10.

```
drupal.js?v=10.1.2:64 Uncaught TypeError: $(...).once is not a function
    at Object.attach (header-menu.js?rzsw7a:16:50)
    at drupal.js?v=10.1.2:166:24
    at Array.forEach (<anonymous>)
    at Drupal.attachBehaviors (drupal.js?v=10.1.2:162:34)
    at big_pipe.js?v=10.1.2:118:10
    at big_pipe.js?v=10.1.2:138:3
```

It looks like several JS files in `server_theme` still have a dependency to the removed library `core/jquery.once`, they were not updated in the [Drupal 10 MR](https://github.com/Gizra/drupal-starter/pull/539):

```
web/themes/custom/server_theme/src/js/accordion.js
web/themes/custom/server_theme/src/js/header-menu.js
web/themes/custom/server_theme/src/js/search.js
web/themes/custom/server_theme/src/js/view-more.js
```

See:
- https://www.gizra.com/content/upgrading-drupal9-to-drupal10/
- https://www.drupal.org/node/3158256

amitaibu:
@bmunslow Thanks

@mariano-dagostino ☝️

amitaibu:
@bmunslow out of curiosity, are you using the repo on your projects?

GizraNotifier:
The latest merged PR #550 just got deployed successfully to Pantheon [`qa`](https://gizra-drupal-starter-qa.pantheonsite.io) environment

amitaibu:
> just got deployed successfully to Pantheon [qa](https://gizra-drupal-starter-qa.pantheonsite.io/) environment

@AronNovak The link to`qa` leads to 404

<img width="1041" alt="image" src="https://github.com/Gizra/drupal-starter/assets/125707/3e0ef077-47d0-4edc-a5c4-acacf45124a7">

GizraNotifier:
The latest merged PR #564 just got deployed successfully to Pantheon [`qa`](https://qa-gizra-drupal-starter.pantheonsite.io) environment

PR #550, opened 2023-08-22
Replace jquery.once implementation with once

mariano-dagostino:
#549 

https://github.com/Gizra/drupal-starter/assets/5313631/9dc9b09b-f910-48a1-862c-581c896c3f90

### DIFF
--- a/web/themes/custom/server_theme/src/js/accordion.js
+++ b/web/themes/custom/server_theme/src/js/accordion.js
@@ -7,7 +7,7 @@
   Drupal.behaviors.accordion = {
     attach: function (context, settings) {
 
-      $('.accordion .accordion-title').once().click(function() {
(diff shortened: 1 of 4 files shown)


amitaibu (approved):
Thanks 👍

web/themes/custom/server_theme/src/js/accordion.js
```diff
@@ -7,7 +7,7 @@
   Drupal.behaviors.accordion = {
     attach: function (context, settings) {
 
-      $('.accordion .accordion-title').once().click(function() {
+      $(once('accordion-toogle', '.accordion .accordion-title')).click(function() {
```
AronNovak, line 10: AFAIK we need to inject once explicitly above & below

mariano-dagostino: @AronNovak No, it seems this is optional as once is a global function.

Issue #551, opened 2023-08-24
Fatal error thrown when creating a Hero Image paragraph with empty field_link


bmunslow:
Steps to reproduce:

- Create new Landing Page
- Add Hero Image paragraph
- Fill in mandatory fields (Title and Image) but leave Link field empty
- Save

https://github.com/Gizra/drupal-starter/assets/295192/5fb21980-698c-48e8-9116-d32c27546813

bmunslow:
Error output:

```
The website encountered an unexpected error. Please try again later.

TypeError: Drupal\server_general\Plugin\EntityViewBuilder\ParagraphHeroImage::buildElementHeroImage(): Argument #4 ($button_text) must be of type string, null given, called in /var/www/html/web/modules/custom/server_general/src/Plugin/EntityViewBuilder/ParagraphHeroImage.php on line 47 in Drupal\server_general\Plugin\EntityViewBuilder\ParagraphHeroImage->buildElementHeroImage() (line 80 of modules/custom/server_general/src/ElementTrait.php).
Drupal\server_general\Plugin\EntityViewBuilder\ParagraphHeroImage->buildFull(Array, Object) (Line: 100)
Drupal\pluggable_entity_view_builder\EntityViewBuilderPluginAbstract->build(Array, Object, 'full') (Line: 147)
Drupal\pluggable_entity_view_builder\ParagraphViewBuilder->doBuild(Array) (Line: 66)
Drupal\pluggable_entity_view_builder\ParagraphViewBuilder->doBuildMultiple(Array) (Line: 37)
Drupal\pluggable_entity_view_builder\ParagraphViewBuilder->buildMultiple(Array) (Line: 238)
Drupal\Core\Entity\EntityViewBuilder->build(Array)
```

amitaibu:
@bmunslow Thanks. Do you want to have a shot at writing a PR for this? 😸 We should keep the Link field optional, but fix the WSOD.

GizraNotifier:
The latest merged PR #554 just got deployed successfully to Pantheon [`qa`](https://gizra-drupal-starter-qa.pantheonsite.io) environment

PR #554, opened 2023-08-24
Add support for empty field_link in paragraph.hero_image

bmunslow:
#551 

A fatal error was thrown when the optional field `field_link` was left empty in `paragraph.hero_image`.

This commit fixes the issue by conditionally rendering the button in the helper method `ElementTrait::buildElementHeroImage`.


bmunslow:
image

web/modules/custom/server_general/src/ElementTrait.php
```diff
@@ -69,15 +69,17 @@ protected function buildElementCta(string $title, array $body, string $button_te
...
+   *   If NULL, no button is rendered. Defaults to NULL.
+   * @param \Drupal\Core\Url|null $url
    *   The URL to link the button to.
+   *   If NULL, no button is rendered. Defaults to NULL.
    *
    * @return array
    *   Render array.
    */
-  protected function buildElementHeroImage(array $image, string $title, string $subtitle, string $button_text, Url $url): array {
+  protected function buildElementHeroImage(array $image, string $title, string $subtitle, string $button_text = NULL, Url $url = NULL): array {
```
amitaibu, line 82: How about changing this to `Link $link`. This means we won't have weird cases where we have button text, but no button URL.

```suggestion
  protected function buildElementHeroImage(array $image, string $title, string $subtitle, Link $link): array {
```

amitaibu (commented):
Thanks! Added a comment 😄

web/modules/custom/server_general/src/ElementTrait.php
bmunslow: Ok, I have amended the commit, please take another look. Notice the `Link` object must have a default `NULL` value.

Also updated the `StyleGuideController` to reflect the change in the parameters passed on to `ElementTrait::buildElementHeroImage`.

amitaibu:
Unfortunately, Travis CI doesn't run on forks, so if you can, please execute manually and attach a screenshot confirming it passes:

- `ddev phpcs` (It should fix a few issues currently in the code)
- `ddev phpstan`
- `ddev phpunit`

bmunslow:
> Unfortunately, Travis CI doesn't run on forks, so if you can, please execute manually and attach a screenshot confirming it passes:
> 
> * `ddev phpcs` (It should fix a few issues currently in the code)
> * `ddev phpstan`
> * `ddev phpunit`

Sure thing!

I amended the commit again with the changes fixed by `ddev phpcs`, it alphabetically ordered `Use` statements, thanks for that.

`ddev phpstan `and `ddev phpunit` also pass cleanly:

image
image
image
image
image
image
image

amitaibu (approved):
👍  Thanks

Issue #553, opened 2023-08-24
Fix wrong background-repeat on Status messages [1h]


amitaibu:
1. Create a node
1. See the status message:

<img width="1351" alt="image" src="https://github.com/Gizra/drupal-starter/assets/125707/af1f915d-8d2a-4a31-bfce-5810e37b90df">

GizraNotifier:
The latest merged PR #556 just got deployed successfully to Pantheon [`qa`](https://gizra-drupal-starter-qa.pantheonsite.io) environment

PR #556, opened 2023-08-25
Fix wrong background-repeat on Status messages.

dipakmdhrm:
#553 

### Update:
- Fixed wrong background-repeat on Status messages.

Before|After
-|-
before|after

**TB:0.5/1h**

### DIFF
--- a/robo-components/ElasticSearchTrait.php
+++ b/robo-components/ElasticSearchTrait.php
@@ -77,7 +77,7 @@ trait ElasticSearchTrait {
    *   The password of the ES admin user.
    * @param string|null $environment
    *   The environment ID. To test changes in the index config selectively.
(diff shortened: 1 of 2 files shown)


robo-components/ElasticSearchTrait.php
```diff
@@ -77,7 +77,7 @@ protected function randomStr(
    *   The password of the ES admin user.
    * @param string|null $environment
    *   The environment ID. To test changes in the index config selectively.
-   * @param bool $needs_users_override
+   * @param bool|null $needs_users_override
```
dipakmdhrm, line 80: Found this on `ddev phpstan`

amitaibu (approved):
Thanks

Issue #552, opened 2023-08-24
Error on `/admin/config` page [3h]


amitaibu:
https://drupal-starter.ddev.site:4443/admin/config

<img width="1369" alt="image" src="https://github.com/Gizra/drupal-starter/assets/125707/8f509062-a532-41d4-b37d-65e1917d9ceb">

PR #560, opened 2023-08-29
Fix strstr error in config path

mariano-dagostino:
#552 

I traced this error to this issue. https://www.drupal.org/project/drupal/issues/3318180 It seems the config missing added by this PR fixes the problem.

@amitaibu I thought in adding a test for this but maybe it will be problematic in a future. checking if there is some message-error in the config page. What do you think?

1/3

### DIFF
--- a/config/sync/locale.settings.yml
+++ b/config/sync/locale.settings.yml
@@ -1,2 +1,7 @@
 translation:
+  default_filename: '%project-%version.%language.po'
+  overwrite_customized: false
+  overwrite_not_customized: false
   path: sites/default/files/translations
+  update_interval_days: 0
+  use_source: local
--- a/web/modules/custom/server_general/tests/src/ExistingSite/ServerGeneralConfigTest.php
+++ b/web/modules/custom/server_general/tests/src/ExistingSite/ServerGeneralConfigTest.php
@@ -1,0 +1,28 @@
+<?php
+
+namespace Drupal\Tests\server_general\ExistingSite;
+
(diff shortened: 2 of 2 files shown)


config/sync/locale.settings.yml
```diff
@@ -1,2 +1,7 @@
 translation:
+  default_filename: '%project-%version.%language.po'
+  overwrite_customized: false
+  overwrite_not_customized: false
```
amitaibu, line 4: Did you add these changes manually? If so, how did you know what to add?

web/modules/custom/server_general/tests/src/ExistingSite/ServerGeneralConfigTest.php
```diff
@@ -0,0 +1,28 @@
...
+
+namespace Drupal\Tests\server_general\ExistingSite;
+
+use Symfony\Component\HttpFoundation\Response;
+use weitzman\DrupalTestTraits\ExistingSiteBase;
+
+/**
+ * Config based tests.
+ */
+class ServerGeneralConfigTest extends ExistingSiteBase {
```
amitaibu, line 11: Nice idea👌

amitaibu (approved):

config/sync/locale.settings.yml
mariano-dagostino: @amitaibu This is a combination of choosing some defaults in the UI and exporting the config and copy from the core `locale.settings.yml` config the other values.

PR #561, opened 2023-08-30
Retry `install_ddev.sh` step

AronNovak:
When we install DDEV, a number of things happen: 
 - Downloading DDEV itself
 - Downloading Docker images from multiple sources sometimes
 - Downloading APT packages

and so one. One network failure should not halt the process.


amitaibu (approved):
👌

PR #565, opened 2023-08-31
Add svgo config to disable stripping viewBox attribute

bboro:
Looks like svgo is responsible for stripping out viewbox attribute which makes scaling worse. Adding this config file disables it.

@see https://github.com/svg/svgo#troubleshooting
and https://github.com/svg/svgo/issues/1128


amitaibu (approved):
👍

Issue #568, opened 2023-09-05
Distribute Robo commands via Packagist


AronNovak:
We have some client projects where we compose custom code via composer. It has a number of benefits:
 - versioned properly
 - easy to upgrade / downgrade
 - coherent with other contrib code

We have this thing that we copy-paste Robo commands from Drupal Starter. From some projects, we diverge so much from the origin that we can't do better. But many projects are really really similar to Starter when it comes to tooling, but yet, each one is a little bit different.

If we'd try to publish a composer package that holds our commands, we would be able to upgrade (let's say to have the latest release notes generation or fixes in the deploy flow) via composer, just like as we upgrade Drupal core.

AronNovak:
^^ @amitaibu Worth to try?

AronNovak:
https://chat.openai.com/share/0ba365e8-13dc-4282-9f02-ac7b591b4034

The local RoboFile would look like:
```
class RoboFile extends MyRoboTasks
{
    // local tasks can also be added here
}

amitaibu:
Sounds like a great idea. And a great demo for AI assistance 👌

AronNovak:
https://github.com/Gizra/robo-release-notes

AronNovak:
https://packagist.org/packages/gizra/robo-release-notes

AronNovak:
https://github.com/Gizra/robo-phpcs

AronNovak:
We have all the packages here: https://packagist.org/packages/gizra/

PR #572, opened 2023-09-21
Refactor Robo components into packagist packages

AronNovak:
#568


RoboFile.php
```diff
@@ -3,16 +3,14 @@
 declare(strict_types = 1);
 
 use Drupal\Core\DrupalKernel;
+use GizraRobo\PhpcsTasks;
```
amitaibu, line 6: Can we have the \name be `use Gizra\RoboPhpcsTasks;`?

PR #575, opened 2023-10-06
Verify facet configuration

AronNovak:
With the Search paragraph type, we need to make sure facet is configured properly, otherwise someone might debug it for a while why facets are missing.


amitaibu (approved):
Cool, thanks

PR #576, opened 2023-10-11
Exclude paths from search index

AronNovak:
To avoid index `/search` itself for instance


mgurjanov:
@AronNovak Maybe you could add info about this to README at https://github.com/Gizra/drupal-starter#elasticsearch.

bboro:
Tests failing and needs reroll @AronNovak

PR #577, opened 2023-10-23
Fix the createMediaImage() function to properly create image media.

dipakmdhrm:
### Update:

- Update `ServerGeneralParagraphTestBase::createMediaImage()` to use `MediaCreationTrait::createMedia` instead of `Media::create()`
- Previously, instead of an image file getting created, a file with uri was being created. Replaced this with an actual image file being created.
- Previously, the file was being attached to `field_media_file` for an `image` media. Image media doesn't have the `field_media_file` field. (I have no idea why this wasn't giving some kind of error). I replaced this with adding image file to `field_media_image` field.


amitaibu:
@dipakmdhrm Thanks. What is the problem this PR is solving?

dipakmdhrm:
> @dipakmdhrm Thanks. What is the problem this PR is solving?

@amitaibu Updated [description](https://github.com/Gizra/drupal-starter/pull/577#issue-1956906536) above.

amitaibu (approved):
Thanks 🙏

PR #579, opened 2023-10-30
PHP 8.2

AronNovak:


AronNovak:
This is an easy upgrade that we might push on client projects too, when we want.

amitaibu (approved):
👍

PR #581, opened 2023-11-06
Fix menu caching

AronNovak:
We need cache tags + contexts for the menu in the header.

### DIFF
--- a/web/modules/custom/server_general/tests/src/ExistingSite/ServerGeneralCacheTest.php
+++ b/web/modules/custom/server_general/tests/src/ExistingSite/ServerGeneralCacheTest.php
@@ -51,7 +51,7 @@ class ServerGeneralCacheTest extends ServerGeneralTestBase {
   }
 
   /**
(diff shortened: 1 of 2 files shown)


web/modules/custom/server_general/tests/src/ExistingSite/ServerGeneralCacheTest.php
```diff
@@ -38,4 +39,44 @@ protected function checkViewForSearchApiTagBasedCaching(View $view) {
...
+  public function testHomepageMenuCache() {
+    $this->drupalGet('<front>');
+    $this->assertMenuState('What We Do');
+    $this->addMenuItem('New Menu Item', 'node/1');
+    $this->drupalGet('<front>');
+    $this->assertMenuState('New Menu Item');
+  }
+
+  /**
+   * Assert the presence of a menu link in the homepage menu.
```
AronNovak, line 54:
```suggestion
   * Assert the presence of a menu link.
```

amitaibu (approved):
Thanks

PR #597, opened 2023-11-24
Search preserve facets

bboro:
Ported from another project.


bboro:
Lots of unrelated file changes, so check the first two commits for related changes.

PR #615, opened 2024-01-09
Add -p flag to mkdir command in config.local.yaml.example

bmunslow:
This prevents mkdir error when the private files folder already exists

`mkdir: cannot create directory ‘/var/www/private’: File exists`


amitaibu (approved):
Thanks

PR #616, opened 2024-01-09
Update Drupal 10.2.1

bmunslow:
This commit updates Drupal to 10.2.1. Notice Drush has also been updated to 12.4 [as required](https://www.drupal.org/docs/updating-drupal/updating-drupal-core-via-composer#s-updating-drupal-core-with-composer).

image
image
image
image
image
image
image


AronNovak (approved):

AronNovak:
Likely Travis is not executed as we cannot use secrets in forks for public repositories.
Otherwise looks good.

PR #617, opened 2024-01-12
Separate Test & Deploy steps.

dipakmdhrm:
### Context:
Sometimes during deployment, the build passes all tests but the deployment fails due to some issue in git, redis, etc. In such cases, to trigger the deployment, we need to re-run the whole test+deploy. 

This will allow us to trigger deployment without triggering tests.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -41,41 +41,37 @@ jobs:
     - "$TRAVIS_BUILD_DIR/ci-scripts/test_shell.sh || travis_terminate 1;"
   - stage: Test
     name: 'Backend tests: Functional tests'
-    if: (branch != "main" AND tag IS blank)
     script:
     - "(travis_retry $TRAVIS_BUILD_DIR/ci-scripts/install_ddev.sh) || travis_terminate 1;"
     - "$TRAVIS_BUILD_DIR/ci-scripts/install_drupal.sh || travis_terminate 1;"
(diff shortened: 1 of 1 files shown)


.travis.yml
```diff
@@ -41,41 +41,37 @@ jobs:
...
     - "ddev composer install || travis_terminate 1;"
     - "(travis_retry ddev robo deploy:pantheon-sync) || travis_terminate 1;"
   - stage: Deploy
-    name: 'Backend tests: Functional tests and deploy to Pantheon LIVE'
+    name: 'Backend tests: Deploy to Pantheon LIVE'
     if: tag IS present AND type = "push" AND tag =~ /live$/
     script:
     - "$TRAVIS_BUILD_DIR/ci-scripts/check_live_deploy.sh || travis_terminate 1;"
     - "(travis_retry $TRAVIS_BUILD_DIR/ci-scripts/install_ddev.sh) || travis_terminate 1;"
     - "$TRAVIS_BUILD_DIR/ci-scripts/install_drupal.sh || travis_terminate 1;"
```
bboro, line 74: Do we still need to install drupal if we're not running tests?

AronNovak:
It was an optimization idea to save time. Initializing the project takes time (downloading Docker images, etc).
I agree that in the bad case, it's the opposite effect, but anyways we need to execute tests for the sake of safety upon deploy, if we bundle it with deploy step, for the good scenario, we save time, I am not entirely convinced that we'd want this.

Issue #614, opened 2024-01-08
Give Elements better names, and seperate into better named traits [6h]


amitaibu:
Right now, we have many elements crammed together under a few traits.

For example, https://github.com/Gizra/drupal-starter/blob/0069ac59f6d88f73fc77021032723c56d227144d/web/modules/custom/server_general/src/ElementTrait.php#L169 lives under `ElementTrait.php,` but could happily live in `CarouselTrait.php`

This one https://github.com/Gizra/drupal-starter/blob/0069ac59f6d88f73fc77021032723c56d227144d/web/modules/custom/server_general/src/ElementTrait.php#L248 can live under `NewsTrait.php` (or `TeaserTrait`?)

In other words, we need to keep the functionality but reduce the abstractions of the names.

So it's not just a matter of dev, but also some dev psychology -- "where would another dev be looking for this element"?

- [ElementTrait.php](https://github.com/Gizra/drupal-starter/blob/0069ac59f6d88f73fc77021032723c56d227144d/web/modules/custom/server_general/src/ElementTrait.php)
- [InnerElementTrait.php](https://github.com/Gizra/drupal-starter/blob/0069ac59f6d88f73fc77021032723c56d227144d/web/modules/custom/server_general/src/InnerElementTrait.php)
- ... and maybe other traits would benefit from this

bboro:
@amitaibu this is now done I suppose. And can be closed if you want to.

GizraNotifier:
The site could not be fully updated at Pantheon at qa. Try fixing manually.

PR #618, opened 2024-01-12
Traits refactoring to more specific names instead of ElementTrait and InnerElementTrait

bboro:
#614 [6h]


bboro:
I think I prefer the latter, if we have an "InfoCard" paragraph type, and "InfoCardTrait", it makes sense that stuff for this paragraph type belongs there, and that's where we should start looking for things. What do you think?

bboro:
3 / 6h will continue Monday. Think I'm heading in the right direction...

amitaibu:
> should break it down further into "InfoCardTrait", "PeopleTrait" etc?

Yes, I agree we should break it down into smaller traits that are less abstract.

web/modules/custom/server_general/src/ElementLayoutTrait.php
```diff
@@ -17,6 +19,7 @@
  */
 trait ElementLayoutTrait {
```
amitaibu, line 20: I think this file is still abstracted. Maybe break it down as well?

web/modules/custom/server_general/src/InnerElementLayoutTrait.php
```diff
@@ -8,15 +8,16 @@
 use Drupal\Core\Url;
 
 /**
- * Helper methods for rendering different Card layouts.
+ * Helper methods for rendering different "inner element" layouts such as cards.
  *
- * A card layout can be for example a card with an image, or a card with
- * centered items. This trait should only be used by `InnerElementTrait`.
+ * An inner element can be for example a card with an image, or a search result
```
amitaibu, line 13: 👍

amitaibu (approved):
Looks great, just a couple of comments, and MWG

Bumnp 1h to address

bboro:
Summary of changes:

- Break down ElementTrait and InnerElementTrait into more specific traits under `Drupal\server_general\ElementTrait\` namespace such as `CtaTrait` and `CarouselTrait`.
- Renamed `buildInnerElement...` to `buildElement...` to keep it simpler
- Moved some more "layout-y" methods from  `ElementTrait` into `ElementLayoutTrait`
- Renamed `server_theme_inner_element__...` themes to `server_theme_element__...` themes to keep it simpler.

We still have some separation of "Element" and "InnerElement" though. The idea of "InnerElement" still exists in `InnerElementLayoutTrait` and contains some layout methods for Cards etc. I decided to keep this structuring separation for now, but we can also perhaps move this to `ElementLayoutTrait` later if needed. For now I updated the comment.

TB: 5 / 6h

amitaibu:
> I decided to keep this structuring separation for now, but we can also perhaps move this to ElementLayoutTrait later if needed. For now I updated the comment.

Yeah, I think the fact you said it's a "private" trait is a good balance 👍

web/modules/custom/server_general/src/ElementLayoutTrait.php
bboro: I left it as it contains just "Layout" stuff now, such as TitleBodyAndItems etc. Where would you move these?

bboro: I'm merging as is for now, we can break this down further later if you want to.

PR #619, opened 2024-01-12
CKEeditor HTML source breaks page layout when there are very long strings

mgurjanov:
When using Claro theme, CKEditor HTML source view breaks page layout when there are very long strings in HTML source (e.g. href value).

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -134,6 +134,9 @@
             "drupal/core": "-p2"
         },
         "patches": {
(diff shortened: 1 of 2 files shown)


bboro (approved):

mgurjanov:
@bboro Please wait with this as it might make Edit view too tall.

mgurjanov:
It was false alarm. Fix is working OK.

PR #620, opened 2024-01-13
Fix translator role issues

mgurjanov:
When we have a translator role that doesn't have edit permissions we encounter two translation issues that this PR fixes:

1. Referenced Paragraph cannot be accessed for translation
2. Media library introduced edit permission check that pure translator role shouldn't have (such translator role is able to only add/update translations and not to edit source node and this is supported by Drupal core). This permission check for such translator role, results in AJAX error which the applied fix prevents and allows a media entity to be changed/translated in translated node.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -134,6 +134,9 @@
             "drupal/core": "-p2"
         },
         "patches": {
+            "drupal/core": {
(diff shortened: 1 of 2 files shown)


amitaibu:
Thanks. Can you please add test(s) to validate?

Issue #626, opened 2024-02-28
Add caching to hook_entity_access()


balagan73:
We could add caching based on the ConfigPages entity used to store locked nodes.

amitaibu:
I'm not sure `hook_entity_access` uses caching.

balagan73:
We have an Access checker in hook_entity_access(), and I am acting based on https://www.drupal.org/docs/8/api/cache-api/access-checkers-cacheability

GizraNotifier:
The site could not be fully updated at Pantheon at qa. Try fixing manually.

PR #627, opened 2024-02-28
Added cache tags based on main_settings ConfigPages.

balagan73:
#626 

Added caching to the access checker based on the 'main_settings' ConfigPages entity.


AronNovak (approved):
Looks good.

Issue #634, opened 2024-03-04
Add rollbar integration


balagan73:
- Add rollbar configuration to settings.pantheon.php

GizraNotifier:
The site could not be fully updated at Pantheon at qa. Try fixing manually.

PR #635, opened 2024-03-04
Added rollbar configuration

balagan73:
#634 

- Added rollbar configuration

### DIFF
--- a/README.md
+++ b/README.md
@@ -213,13 +213,11 @@ and you can see the test running in the browser.
 Check the [DDEV documentation](https://ddev.readthedocs.io/en/latest/users/debugging-profiling/step-debugging/)
 if you are using other IDE or want to know more about this feature.
 
-## Prepare for Deployment
+## Deploy to Pantheon
 
 ### Rollbar configuration
 
-Rollbar will be automatically enabled on test and live environments. The access tokens need to be added and the rollbar configuration yml file should be committed.
-
-## Deploy to Pantheon
+Rollbar will be automatically enabled on test and live environments. The access tokens need to be added under /admin/config/services/rollbar and the rollbar configuration yml file should be committed.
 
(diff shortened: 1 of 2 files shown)


web/sites/default/settings.pantheon.php
```diff
@@ -70,9 +70,12 @@
     }
   }
 }
-
-$pantheon_env = getenv('PANTHEON_ENVIRONMENT');
 if (!empty($pantheon_env)) {
+  // Rollbar settings for LIVE and TEST.
+  if ($pantheon_env == 'live' || $pantheon_env == 'test') {
+    $config['rollbar.environment'] = $pantheon_site_name . '.' . $pantheon_env;
+    $config['rollbar.enabled'] = TRUE;
```
amitaibu, line 77: Here we would also have the access token of rollbar right? If so, let's add a placeholder

balagan73: No. Access token is saved in the module's config file, but there is a setting which disables the module, here we just enable it.

amitaibu (commented):
Let's add a note in the README, about settings Rollbar.

Maybe under `## Prepare for Deploy` section or similar

README.md
```diff
@@ -215,6 +215,10 @@ if you are using other IDE or want to know more about this feature.
 
 ## Deploy to Pantheon
 
+### Rollbar configuration
+
+Rollbar will be automatically enabled on test and live environments. The access tokens need to be added under '/admin/config/services/rollbar' and the rollbar configuration yml file should be committed.
```
amitaibu, line 220: Actually, instead of `/admin/config/services/rollbar` - can't we have the setting ready on `settings.pantheon.php`?

AronNovak (approved):

Issue #638, opened 2024-03-08
Fix Dynamic Page Caching


mariano-dagostino:
There is something preventing Dynamic Page Cache working on the homepage and any page of drupal-starter.

![Screenshot_2024-03-07_15-21](https://github.com/Gizra/drupal-starter/assets/5313631/9cdc4b53-becf-47b5-9a07-1d84d57e65ac)

GizraNotifier:
The site could not be fully updated at Pantheon at qa. Try fixing manually.

PR #639, opened 2024-03-08
Use block_plugin_view_builder module to render blocks. Provide cache_review as module dev dependency. Display header information when using ddev

mariano-dagostino:
#638

This PR changes the way drupal starter renders blocks. By using [block_plugin_view_builder](https://www.drupal.org/project/block_plugin_view_builder) instead of the previous method, drupal can delay the rendering of some blocks and make proper use of bigpipe.

A bit more of context about why this is important. When rendering the Language Switcher block, drupal sees a Max-Age of 0 and decides the page cannot be cached. https://www.drupal.org/node/2232375#comment-8647129 When the previous method of rendering blocks, the max-age was bubbled up to the entire page and therefore marked as UNCACHEABLE by Dynamic Page Cache. By using block_plugin_view_builder module the block is now a lazy_builder callback and drupal can make better decisions of what to cache and what to not cache.

At first glance this provides a 50% page speed improvement for authenticated users. With the current codebase of drupal-starter.

Also  it enables the debugging of headers when using ddev locally. Something that was not enabled by default.

Finally it adds [cache_review](https://www.drupal.org/project/cache_review) module as a dev dependency which proved to be handy to debug this problem.

Before|After
-|-
Screenshot_2024-03-07_15-21|Screenshot_2024-03-08_10-25
Screenshot_2024-03-08_10-16_1|Screenshot_2024-03-08_10-17
Screenshot_2024-03-08_10-16|Screenshot_2024-03-08_10-17_1


AronNovak (approved):
This is a wonderful finding!

amitaibu (approved):
Great :bow:

Issue #641, opened 2024-03-14
When trying to delete locked pages, access should be denied instead of exception


balagan73:
> /delete should give an access denied, not an Exception

GizraNotifier:
The site could not be fully updated at Pantheon at qa. Try fixing manually.

PR #642, opened 2024-03-14
When trying to delete locked pages, access should be denied instead of exception

balagan73:
#641 

- Displaying error when user tries to delete locked page

TB: 1h


balagan73:
image

amitaibu:
The `Delete` tab should not appear on locked pages.

amitaibu:
and trying to go to `/delete` should already give a 403

balagan73:
One step from being finished. Now after removing static Drupal:: calls and using DI, getting this error
image

balagan73:
Locked page|Not locked page
-|-
image|image

web/modules/custom/server_general/src/Routing/LockedPagesRouteSubscriber.php
```diff
@@ -0,0 +1,119 @@
...
+    }
+  }
+
+  /**
+   * Returns the 'main_settings' config page.
+   *
+   * @return \Drupal\Core\Entity\ContentEntityInterface|null
+   *   The 'main_settings' config page entity or null if not found.
+   */
+  protected function getMainSettings() {
```
amitaibu, line 62: Most of the code in this class is copied from another class? If so, please use a trait instead

amitaibu:
Let's add tests to verify this. Bump 1h

web/modules/custom/server_general/src/Routing/LockedPagesRouteSubscriber.php
balagan73: It is copied from the .module file. Yeah, trait is a good idea.

balagan73: Instead of using trait, I plan to create another service encapsulating the common functions used both here and in the module file.

balagan73:
Admin user deleting locked page|Anonymous trying to delete a landing page|Content editor deleting own page
-|-|-
image|image|image

amitaibu:
> Content editor deleting own page

It should also prevent deletion. If the page is locked, then deletion can't happen

andreytroeglazov:
I think if node is locked, it can't be deleted by any action, even if you have permissions or you are the author.

Also maybe we need remove all instances of `landing_page` so it works for any locked node not depending on the content type.

And the last point, I think all this functionality can be moved to contrib module and published on d.o., I will do it when I have some free time :thinking:

balagan73:
@amitaibu Just tested that user can delete own nodes, if those are not locked. Locked nodes cannot be deleted at all, not even programmatically.

amitaibu:
> Also maybe we need remove all instances of landing_page so it works for any locked node not depending on the content type

We can do that.

>  I think all this functionality can be moved to contrib module and published on d.o

Sounds good

web/modules/custom/server_general/src/Routing/LockedPagesRouteSubscriber.php
```diff
@@ -0,0 +1,98 @@
...
+    // Return in_array($node->id(), $restricted_entities);.
+    if ($node instanceof NodeInterface && in_array($node->getType(), $bundles)) {
+      $main_settings = $this->lockedPagesService->getMainSettings();
+      $cache_tags = $main_settings->getCacheTags();
+      if ($this->lockedPagesService->isNodeLocked($node)) {
+        return AccessResult::forbidden()->addCacheableDependency($node)->addCacheTags($cache_tags);
+      }
+    }
+    // Allow deletion if user has permission to delete any page.
+    if ($account->hasPermission('delete any {$node->getType()} content')) {
```
andreytroeglazov, line 85: I think this `if` and below should be removed, some content editor or admin with these permissions can open wrong node and remove it by accident, so if they really want to remove it, they will need to remove it from configs first.

andreytroeglazov (approved):
MWG after comment

web/modules/custom/server_general/src/Routing/LockedPagesRouteSubscriber.php
```diff
@@ -0,0 +1,98 @@
...
+      return AccessResult::allowed();
+    }
+
+    // If user is not anonymous, and has permission to delete own
+    // content, allow it.
+    if ($account->id() !== 0 && $node->getOwnerId() === $account->id() && $account->hasPermission('delete own {$node->getType()} content')) {
+      return AccessResult::allowed();
+    }
+
+    return AccessResult::forbidden();
```
andreytroeglazov, line 95: I think this should be neutral()?

web/modules/custom/server_general/src/Routing/LockedPagesRouteSubscriber.php
balagan73: We check locked nodes above these lines, and return AccessResult::forbidden(); These checks only run for nodes not locked. I will remove these, if I can return AccessResult::neutral(), but I have tried that and did not work.

andreytroeglazov: Yes, sorry, I see now, but I think AccessResult::neutral() should work :thinking:

balagan73: If I remove these lines, and return ::neutral(), the deletion of any Landing page is not possible. I think this is because we are not in an access hook, but this is a method solely responsible for access checking.

andreytroeglazov: Let me check

andreytroeglazov: Ok it's bug in core https://www.drupal.org/project/drupal/issues/2991698
if any of access result for route is `neutral()` then it throws 403 error.

andreytroeglazov:
If `$this->drupalLogin($user);` is used second time for the same user it fails because looks like TFA module removes core message about login. :smile:

Issue #644, opened 2024-03-20
Disallow content moderation state changes for locked pages, if module is enabled


balagan73:
If the content_moderation module is enabled, state changes can happen making locked pages unavailable. We should not allow it.

PR #645, opened 2024-03-20
Disallow changes for content moderation state for locked pages

balagan73:
#644 

- Checking if content_moderation module is enabled, then resetting published state, if it was the original.


balagan73:
We don't have content moderation in DS, but as we sync this code from here to projects, where it might be installed, I thought it is worth to add it here.

andreytroeglazov (approved):

web/modules/custom/server_general/server_general.module
```diff
@@ -154,8 +154,17 @@ function server_general_node_presave(NodeInterface $entity) {
     return;
   }
 
+  // Get the module handler service.
+  $module_handler = \Drupal::service('module_handler');
   /** @var \Drupal\node\NodeInterface $original */
   $original = $entity->original;
+  // If content moderation is used, don't allow status change for locked pages.
+  if ($module_handler->moduleExists('content_moderation')) {
+    if ($original->get('moderation_state')->getString() === 'published' && $entity->get('moderation_state')->getString() !== 'published') {
```
amitaibu, line 163: Can be in a single IF

web/modules/custom/server_general/server_general.module
```diff
@@ -154,8 +154,15 @@ function server_general_node_presave(NodeInterface $entity) {
     return;
   }
 
+  // Get the module handler service.
+  $module_handler = \Drupal::service('module_handler');
   /** @var \Drupal\node\NodeInterface $original */
   $original = $entity->original;
+  // If content moderation is used, don't allow unpublish for locked pages.
+  if ($module_handler->moduleExists('content_moderation') && $original->get('moderation_state')->getString() === 'published' && $entity->get('moderation_state')->getString() === 'unpublished') {
```
amitaibu, line 162: Maybe better?

```suggestion
  if ($module_handler->moduleExists('content_moderation') && $original->get('moderation_state')->getString() === 'published' && $entity->get('moderation_state')->getString() !== 'published') {
```

balagan73: In another project the specific request was to only disallow unpublishing, as draft etc creates new revisions. I think it makes sense.

balagan73:
I have created a module with the name server_locked_pages, to help syncing changes related to locked pages.

Delete option is not displayed|Delete path cannot be accessed
-|-
image|image

amitaibu:
> I have created a module with the name server_locked_pages

Let's keep it inside server_general. Later, we'll move it out into a contrib module

web/modules/custom/server_general/server_general.module
balagan73: @amitaibu This is the only question now, whether we should disallow only unpublish or the other states too. As I noted above, I think disallowing only unpublish is fine.

amitaibu: >  As I noted above, I think disallowing only unpublish is fine.

:+1:

Issue #647, opened 2024-03-25
Enable Content moderation [2h]


amitaibu:
States:

- Draft
- Review
- Published

For role:

- Editor

balagan73:
@amitaibu Don't we need an 'Unpublished' state? I guess 'Review' should be called 'Ready for review'.

amitaibu:
Sure, go ahead

GizraNotifier:
The site could not be fully updated at Pantheon at qa. Try fixing manually.

PR #648, opened 2024-03-25
Enable content moderation

balagan73:
#647 

- Enabled content moderation module
- Added Default workflow with transitions and states.

TB: 1.5/2h

### DIFF
--- a/web/modules/custom/server_default_content/content/file/3d348c2e-091c-4cd9-a6f5-fe2040030871.yml
+++ b/web/modules/custom/server_default_content/content/file/3d348c2e-091c-4cd9-a6f5-fe2040030871.yml
@@ -4,12 +4,15 @@ _meta:
   uuid: 3d348c2e-091c-4cd9-a6f5-fe2040030871
   default_langcode: en
 default:
+  uid:
+    -
+      target_id: 1
   filename:
     -
       value: seaside-850x625_55.jpg
   uri:
     -
-      value: 'public://seaside-850x625_55.jpg'
+      value: 'public://seaside-850x625_55_12.jpg'
   filemime:
     -
       value: image/jpeg
--- a/web/modules/custom/server_default_content/content/file/3d348c2e-091c-4cd9-a6f5-fe2040030871.yml
+++ b/web/modules/custom/server_default_content/content/file/3d348c2e-091c-4cd9-a6f5-fe2040030871.yml
@@ -4,15 +4,12 @@ _meta:
   uuid: 3d348c2e-091c-4cd9-a6f5-fe2040030871
   default_langcode: en
 default:
-  uid:
-    -
-      target_id: 1
   filename:
     -
       value: seaside-850x625_55.jpg
   uri:
     -
-      value: 'public://seaside-850x625_55_12.jpg'
+      value: 'public://seaside-850x625_55.jpg'
   filemime:
     -
       value: image/jpeg
--- a/web/modules/custom/server_default_content/content/file/3d348c2e-091c-4cd9-a6f5-fe2040030871.yml
+++ b/web/modules/custom/server_default_content/content/file/3d348c2e-091c-4cd9-a6f5-fe2040030871.yml
@@ -12,7 +12,7 @@ default:
       value: seaside-850x625_55.jpg
   uri:
     -
-      value: 'public://seaside-850x625_55_12.jpg'
+      value: 'public://seaside-850x625_55.jpg'
   filemime:
     -
       value: image/jpeg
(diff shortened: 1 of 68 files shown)


balagan73:
Transitions for content editor

From Published|From Draft|From Ready for review
-|-|-
image|image|image

andreytroeglazov:
I think we need `Unpublished` state, because with these 3 currently added editor won't be able to unpublish node completely, instead of this new revision will be created and old published will still be published.

web/modules/custom/server_default_content/content/file/3d348c2e-091c-4cd9-a6f5-fe2040030871.yml
```diff
@@ -4,12 +4,15 @@ _meta:
...
+  uid:
+    -
+      target_id: 1
   filename:
     -
       value: seaside-850x625_55.jpg
   uri:
     -
-      value: 'public://seaside-850x625_55.jpg'
+      value: 'public://seaside-850x625_55_12.jpg'
```
amitaibu, line 15: Why do we have those changes?

balagan73: Not sure. Files are displaying correctly. I just did a new `ddev drush dcem server_default_content` and this changed.

balagan73: At every ddev restart with clean install a new file is created. I will remove these changes.

web/modules/custom/server_default_content/content/file/dc24d0dd-420e-49a4-8ef5-63606fb36767.yml
```diff
@@ -4,12 +4,15 @@ _meta:
...
+  uid:
+    -
+      target_id: 1
   filename:
     -
       value: drupal-starter_10.pdf
   uri:
     -
-      value: 'public://drupal-starter_10.pdf'
+      value: 'public://drupal-starter_10_12.pdf'
```
andreytroeglazov, line 15:
```suggestion
      value: 'public://drupal-starter_10.pdf'
```

andreytroeglazov: Also I think you can revert all changes to default content which are not related to content moderation

web/modules/custom/server_default_content/content/file/3d348c2e-091c-4cd9-a6f5-fe2040030871.yml
```diff
@@ -4,6 +4,9 @@ _meta:
   uuid: 3d348c2e-091c-4cd9-a6f5-fe2040030871
   default_langcode: en
 default:
+  uid:
```
amitaibu, line 7: Why are all those changes?

web/modules/custom/server_general/server_general.module
```diff
@@ -102,6 +102,35 @@ function server_general_field_widget_single_element_language_select_form_alter(a
...
+ * Unpublish option should not be available for moderation of locked nodes.
+ */
+function server_general_field_widget_single_element_moderation_state_permissions_default_form_alter(array &$element, FormStateInterface $form_state, array $context) {
+  /** @var \Drupal\Core\Entity\EntityFormInterface $form_object */
+  $form_object = $form_state->getFormObject();
+  /** @var \Drupal\node\NodeInterface $entity */
+  $entity = $form_object->getEntity();
+
+  if (!$entity instanceof NodeInterface) {
+    return;
```
amitaibu, line 117: Add comment why, here and below

`// Not a node.`

web/modules/custom/server_default_content/content/file/3d348c2e-091c-4cd9-a6f5-fe2040030871.yml
balagan73: Reverted these chanes per the advice of @andreytroeglazov

mgurjanov:
@balagan73 My import failed and this is why I didn't see `Unpublished` state. I fixed it but still think you need to alter transitions. Unpublished state can go only to Unpublished which is wrong, as you want to create draft or ready for review from Unpublished or at least to create Draft. Also you don't have option to make Published->Unpublished. 
Maybe we could also add documentation entry explaining that if node was Published (withing moderation) and is then set to either 'Ready for review' or 'Draft' it won't change its previous core Published status. To be really unpublished it needs to transition to `Unpublished` because only that state is changing its core status value.

balagan73:
I have updated the transitions
image

mgurjanov (approved):
LGTM now. Thanks.

Issue #649, opened 2024-03-25
Fallback to default language if there is no node translation for current language


mgurjanov:
### Issue

There is an issue with Drupal core which prevents fallback to node's default language if there is no translation for current selected language. Default behavior is to fallback to the first translation that exists based on languages alphabetical sort.

### Todo

Fix the issue by applying patch from https://www.drupal.org/project/drupal/issues/3315245

GizraNotifier:
The site could not be fully updated at Pantheon at qa. Try fixing manually.

PR #650, opened 2024-03-25
Fallback to default language if there is no node translation for current language

mgurjanov:
#649 

- [x] fix fallback to default language when there is no translation for current language


amitaibu (approved):
Thanks

PR #654, opened 2024-04-05
Automate some tasks after manual db import

bboro:
Automate some tasks after manual db import. These are some things that I do every time I manually import a db. I think it would be nice to have this automated.


AronNovak (approved):
:+1:

Issue #658, opened 2024-04-16
Bootstrapping fails with Drupal requirement error [1h]


AronNovak:
```
 [ExecStack] Executing terminus remote:drush project.dev -- si server --no-interaction --existing-config
 [ExecStack] Running terminus remote:drush project.dev -- si server --no-interaction --existing-config
 [warning] This environment is in read-only Git mode. If you want to make changes to the codebase of this site (e.g. updating modules or plugins), you will need to toggle into read/write SFTP mode first.
Warning: Permanently added '[appserver.dev.abc-d372-48fb-aec4-def.drush.in]:2222,[34.173.145.32]:2222' (RSA) to the list of known hosts.
 You are about to:
 * DROP all tables in your 'pantheon' database.

 [notice] Starting Drupal installation. This takes a while.
 [notice] Performed install task: install_select_language
 [notice] Performed install task: install_select_profile
 [notice] Performed install task: install_load_profile

In install.core.inc line 2329:
                                                                               
  Settings file:The <em class="placeholder">Settings file</em> is not writabl  
  e.<br /><br />The Drupal installer requires write permissions to <em class=  
  "placeholder">./sites/default/settings.php</em> during the installation pro  
  cess. The <a href="https://www.drupal.org/server-permissions">webhosting is  
  sues</a> documentation section offers help on this and other topics. 
```

Which is unexpected, especially that the installation through the web (`install.php`) goes through.

AronNovak:
```
  Default settings file:The default settings file does not exist.<br /><br />  
  The Drupal installer requires that the <em class="placeholder">./sites/defa  
  ult/default.settings.php</em> file must not be deleted or modified from the  
   original download.
```

PR #659, opened 2024-04-16
Site install process fixes during bootstrap

AronNovak:
#658

### DIFF
--- a/.travis.template.yml
+++ b/.travis.template.yml
@@ -16,6 +16,7 @@ env:
 
 before_install:
   - export PATH="$HOME/.config/composer/vendor/bin:$PATH"
+  - tmpdaemon=$(mktemp)
   - sudo jq '."registry-mirrors" = ["DOCKER_MIRROR"]' /etc/docker/daemon.json > $tmpdaemon
   - sudo mv $tmpdaemon /etc/docker/daemon.json
   - sudo systemctl daemon-reload
--- a/robo-components/DeploymentTrait.php
+++ b/robo-components/DeploymentTrait.php
@@ -562,7 +562,9 @@ trait DeploymentTrait {
     }
 
     $task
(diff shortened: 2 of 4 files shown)


.travis.template.yml
```diff
@@ -16,6 +16,7 @@ env:
 
 before_install:
   - export PATH="$HOME/.config/composer/vendor/bin:$PATH"
+  - tmpdaemon=$(mktemp)
```
amitaibu, line 19: What is it doing?

robo-components/DeploymentTrait.php
```diff
@@ -562,7 +562,9 @@ public function deployPantheonInstallEnv(string $env = 'qa', string $pantheon_na
     }
 
     $task
+      ->exec("terminus connection:set $pantheon_terminus_environment sftp")
```
amitaibu, line 565: Worth a comment

amitaibu (approved):
Thanks

.travis.template.yml
AronNovak: It just uses a temporary file to do manipulations on `/etc/docker/daemon.json` . We can't redirect to the currently manipulated file, as the shell makes it empty as a very first step.

robo-components/DeploymentTrait.php
```diff
@@ -562,7 +562,9 @@ public function deployPantheonInstallEnv(string $env = 'qa', string $pantheon_na
     }
 
     $task
```
AronNovak, line 564:
```suggestion
    // Drupal checks is the settings.php is writable. With the connection mode switch, we can fulfill this.
    $task
```

robo-components/DeploymentTrait.php
AronNovak: Added.

Issue #662, opened 2024-04-23
phpstan deprecation warning [1h]


AronNovak:
```
 aaron   main  ~  gizra  drupal-starter  ddev phpstan
PHP Deprecated:  mb_encoding_aliases(): Handling Base64 via mbstring is deprecated; use base64_encode/base64_decode instead in phar:///var/www/html/vendor/phpstan/phpstan/phpstan.phar/src/Type/Php/MbFunctionsReturnTypeExtension.php on line 41
Deprecated: mb_encoding_aliases(): Handling Base64 via mbstring is deprecated; use base64_encode/base64_decode instead in phar:///var/www/html/vendor/phpstan/phpstan/phpstan.phar/src/Type/Php/MbFunctionsReturnTypeExtension.php on line 41
PHP Deprecated:  mb_encoding_aliases(): Handling Uuencode via mbstring is deprecated; use convert_uuencode/convert_uudecode instead in phar:///var/www/html/vendor/phpstan/phpstan/phpstan.phar/src/Type/Php/MbFunctionsReturnTypeExtension.php on line 41
```

GizraNotifier:
Could not deploy PR #663 to Pantheon properly.

GizraNotifier:
Could not deploy PR #663 to Pantheon properly.

GizraNotifier:
Could not deploy PR #663 to Pantheon properly.

GizraNotifier:
The site could not be fully updated at Pantheon at qa. Try fixing manually.

PR #663, opened 2024-04-23
Upgrade phpstan for PHP 8.2 support

AronNovak:
#662


robo-components/DeploymentTrait.php
```diff
@@ -486,9 +486,6 @@ public function deployCheckRequirementErrors(string $environment): void {
 
     $errors = [];
     $parsed_output = str_getcsv($output, "\n");
-    if (empty($parsed_output)) {
```
AronNovak, line 489: `str_getcsv` returns an array always

andreytroeglazov (approved):

PR #664, opened 2024-04-30
Use experimental navigation module instead of Toolbar and Admin Toolbar

mariano-dagostino:
This PR tries the experimental https://www.drupal.org/project/navigation module.

I think it is interesting to try it because. It would allow uninstall several modules. Also to know better why this module does not work in the default theme_server theme. 

https://github.com/Gizra/drupal-starter/assets/5313631/fb619ac6-b648-4c4d-8e51-e65e8eae2dec

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -56,7 +56,7 @@
         "drupal/migrate_plus": "^6.0",
         "drupal/migrate_source_csv": "^3.4",
         "drupal/migrate_tools": "^6.0",
-        "drupal/navigation": "1.x-dev@dev",
+        "drupal/navigation": "^1.0@beta",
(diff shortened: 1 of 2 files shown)


balagan73 (approved):

amitaibu:
Thanks, looks interesting. How does it behave on Mobile?

amitaibu:
Maybe we should just go with Gin as an Admin theme?

amitaibu:
Yeah, sorry -- I meant I think Gin looks better. But we can keep Claro.
As for the current effort, I'm not sure if it's worth it

amitaibu:
Sure, let's give it 4h

PR #665, opened 2024-04-30
Remove advagg

mariano-dagostino:
It is not installed anymore because of #540  and also

https://www.drupal.org/project/advagg/issues/3308099#comment-15025561


AronNovak (approved):

bboro:
Shouldn't we add it back? Why was this removed in the first place? Can't see anything in the issue

AronNovak:
> I think the advagg module is obsolete for >90% of sites using it in Drupal 10.1! 🤯

PR #666, opened 2024-05-17
TFA- only on live

AronNovak:


amitaibu:
Where do we indicate it should be enabled on LIVE?

AronNovak:
@amitaibu 
Only half of the patch was ported, sorry :smiley_cat:

amitaibu (approved):
Thanks

PR #678, opened 2024-06-21
Upgrade to drupal 10.3. Upgrade contrib modules. Remove old patches. Fix environment_indicator color issue

mariano-dagostino:
Issues to keep an eye on: https://www.drupal.org/project/admin_toolbar/issues/3456128

This PR:

- Upgrades to drupal core 10.3
- Upgrades contrib modules to their last known release.
- Removes the patch introduced in https://github.com/Gizra/drupal-starter/pull/619 as it seems fixed by ckeditor now.
- Fixes a problem with environment indicator displaying white text over white background.

### DIFF
--- a/web/sites/default/settings.ddev.php
+++ b/web/sites/default/settings.ddev.php
@@ -73,7 +73,7 @@ $settings['file_private_path'] = '/var/www/private';
 $settings['config_sync_directory'] = '../config/sync';
 $config['config_split.config_split.dev']['status'] = TRUE;
 $config['environment_indicator.indicator']['bg_color'] = '#00a073';
-$config['environment_indicator.indicator']['fg_color'] = '#000000';
+$config['environment_indicator.indicator']['fg_color'] = '#ffffff';
 
 // SMTP settings. Use Mail Hog (`ddev describe` to get the URL) to see the sent
 // mails.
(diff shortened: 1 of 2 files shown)


AronNovak:
And with this release, we will have taxonomy revisions as well. Nice work @mariano-dagostino

web/sites/default/default.services.yml
```diff
@@ -1,4 +1,8 @@
 parameters:
+  # Toggles the super user access policy. If your website has at least one user
+  # with the Administrator role, it is advised to set this to false. This allows
+  # you to make user 1 a regular user, strengthening the security of your site.
+  security.enable_super_user: true
```
AronNovak, line 5:
```suggestion
  security.enable_super_user: false
```

We already have logic to not use UID 1.

web/sites/default/settings.ddev.php
```diff
@@ -88,8 +88,8 @@
 $settings['container_yamls'][] = 'modules/contrib/redis/redis.services.yml';
 $settings['container_yamls'][] = 'modules/contrib/redis/example.services.yml';
 
-$config['system.performance']['css']['preprocess'] = FALSE;
-$config['system.performance']['js']['preprocess'] = FALSE;
+//$config['system.performance']['css']['preprocess'] = FALSE;
+//$config['system.performance']['js']['preprocess'] = FALSE;
```
AronNovak, line 92: Intentional? It seems out of scope to change it here like this. If intentional, needs comments.

AronNovak (approved):

PR #692, opened 2024-07-19
Fix webp issue 3281606

mariano-dagostino:
https://www.drupal.org/project/webp/issues/3281606

### DIFF
--- a/composer.patches.json
+++ b/composer.patches.json
@@ -28,7 +28,7 @@
       "d.o. #3456599": "https://www.drupal.org/files/issues/2024-06-24/3456599-6.patch"
     },
     "drupal/webp": {
-      "d.o. #3281606": "https://git.drupalcode.org/project/webp/-/merge_requests/33.patch"
+      "d.o. #3281606": "https://git.drupalcode.org/project/webp/-/merge_requests/33.diff"
     },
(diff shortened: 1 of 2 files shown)


composer.patches.json
```diff
@@ -27,6 +27,9 @@
     "drupal/environment_indicator": {
       "d.o. #3456599": "https://www.drupal.org/files/issues/2024-06-24/3456599-6.patch"
     },
+    "drupal/webp": {
+      "d.o. #3281606": "https://git.drupalcode.org/project/webp/-/merge_requests/33.patch"
```
bboro, line 31:
```suggestion
      "d.o. #3281606": "https://git.drupalcode.org/project/webp/-/merge_requests/33.patch"
```

It would be better to save the patch locally I think in case git.drupalcode.org is unavailable.

Also I think it would be better to use the .diff file: https://git.drupalcode.org/project/webp/-/merge_requests/33.diff instead of a .patch. But you can save it locally as a *.patch file.

bboro: I think we can still download modules from drupal.org but git.drupalcode.org is more of a source code hosting. IIRC sometimes it gets stuck asking for a key because of rate limits n stuff. But I might be misremembering.

PR #695, opened 2024-07-30
Activate Git mode upon deploy

AronNovak:
If someone temporarily activates Git mode for debugging, autodeploy will fail in a way which is not really easy to understand or debug. At a client project, deploying manually revealed the root cause, but it did block the deploy flow for a while.


amitaibu (approved):
Thanks

Issue #688, opened 2024-07-15
Document suggested way of changing the default language


bmunslow:
This starter ships with English as the default language and two additional languages to enable translations.

How do you go about it when the project has a single language, other than English? Or perhaps when the project has several languages but none of them is English?

Since changing the default language in Drupal is [no easy job](https://www.drupal.org/node/2796711), I was wondering if this process could be documented in the readme file.

isalmanhaider:
`drush site:install --locale=ar`

or

one solution could be utilising https://www.drupal.org/project/drush_language

```
drush language:add fr
drush language:default fr
```

isalmanhaider:
`drush site:install --locale=ar`

or

one solution could be utilising https://www.drupal.org/project/drush_language

```
drush language:add ar
drush language:default ar
```

bmunslow:
Thanks for the suggestion @isalmanhaider. The only drawback of your solution is it prevents setting up the site with the existing config, which is the whole point of using this Drupal starter.

These are the steps I had to take in a recent project to set Spanish as the default language:

- Start project as usual
- Delete unwanted languages in `/admin/config/regional/language`
- Set desired language as default (in this case, Spanish)
- Export config (`ddev drush cex`)
- Find and replace string "`langcode: en`" with "`langcode: es`" in the following folders:
  - `config/sync`
  - `web/modules/custom/server_default_content`
- Restart project (`ddev restart)`

No issues found for the time being. This approach doesn't take into account existing config for Solr setup which is tailored for English language.

bmunslow:
Hi @mariano-dagostino thanks a lot for looking into this! I tried your script and it seems to work very well on a clean install. Selected language is indeed set as default and all config and content are correctly set. It looks from the comments in the PR, though, that this will remain as a experimental feature for the time being. In any case, it's great to have your script hanging around and I will very likely use it, thanks again!

PR #704, opened 2024-09-04
Provide a way to automate the instalation of drupal in a language other than english

mariano-dagostino:
#688

### DIFF
--- a/.ddev/commands/web/setup-language
+++ b/.ddev/commands/web/setup-language
@@ -2,14 +2,15 @@
 
 ## Description: Configures drupal to install language
 ## Usage: setup-language LANGCODE
-## Example: "ddev setup-language es"
+## Example: "ddev setup-language es Spanish"
 
-if [ $# -ne 1 ]; then
-  echo "Usage: $0 <langcode>"
+if [ $# -ne 2 ]; then
+  echo "Usage: $0 <langcode> <language name>"
   exit 1
 fi
 
 langcode=$1
+langname=$2
 
 # The site is configured in english. Nothing to do here.
 if [ "$langcode" == "en" ]; then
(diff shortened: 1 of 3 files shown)


.ddev/commands/web/setup-language
```diff
@@ -0,0 +1,46 @@
+#!/bin/bash
```
AronNovak, line 1: I see it a bit fragile to do it with string manipulations.
Why not doing this inside Robo?
Then, as Drupal is bootstrapped, we can simply use the API to achieve this. Then we can a `drush cex` and a `git commit` even.

https://gorannikolovski.com/snippet/how-programmatically-add-language#:~:text=Sometimes%20you%20have%20to%20add,)%3B%20%24language%2D%3Esave()%3B

Also 

```
<?php

// Ensure this file is executed in the context of a Drupal bootstrap.
use Drupal\Core\Language\LanguageInterface;
use Drupal\language\Entity\ConfigurableLanguage;

if (!drupal_bootstrap(DRUPAL_BOOTSTRAP_FULL)) {
    drush_log('This script must be run in the context of a fully bootstrapped Drupal site.', 'error');
    exit(1);
}

// Define the language code of the new language to be installed.
$new_language_code = 'es'; // Spanish (change 'es' to the desired language code).

// Check if the language already exists.
$existing_language = \Drupal::languageManager()->getLanguage($new_language_code);

if ($existing_language) {
    drush_log(dt('The language @lang_code already exists.', ['@lang_code' => $new_language_code]), 'warning');
    return;
}

// Create and save the new language.
ConfigurableLanguage::create([
    'id' => $new_language_code,
    'label' => 'Spanish', // Change this to the appropriate language name.
    'weight' => 0,
    'locked' => LanguageInterface::LOCKED
])->save();

drush_log(dt('The language @lang_code has been added.', ['@lang_code' => $new_language_code]), 'success');

// Rebuild the cache.
drupal_flush_all_caches();

drush_log('Cache has been cleared.', 'success');
```

(AI, as-is)

amitaibu:
Thanks. 

I think we should hold this PR, until we get enough real use cases for this (as currently we don't have any in Gizra AFAIK). I think for now whoever tries it could go with the manual work described in https://github.com/Gizra/drupal-starter/issues/688#issuecomment-2293094582

Issue #605, opened 2023-12-06
Add `ddev robo cache:disable|enable` [3h]


amitaibu:
A shorter command to execute https://www.drupal.org/docs/develop/development-tools/disable-caching#s-disable-twig-caching for enable/ disable

PR #705, opened 2024-09-05
Add ddev robo cache:disable|enable

mgurjanov:
#605 

- [x] add new robo cache enable/disable commands

Note: I tested locally and I am getting cache HIT/MISS according to appropriate new command.

Example 1|Example 2|Example 3
-|-|-
ds_cache_enable1|ds_cache_disable|ds_cache_enable1

### DIFF
--- a/RoboFile.php
+++ b/RoboFile.php
@@ -3,6 +3,7 @@
 declare(strict_types=1);
 
 use Drupal\Core\DrupalKernel;
+use Robo\Exception\TaskException;
 use Robo\Tasks;
 use RoboComponents\BootstrapTrait;
 use RoboComponents\DeploymentTrait;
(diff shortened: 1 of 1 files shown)


RoboFile.php
```diff
@@ -71,4 +84,186 @@ public function __construct() {
     }
   }
 
+  /**
+   * Helper to get settings.local.php file.
```
AronNovak, line 88: Put all these into a trait, let's say `DevelopmentTrait` under `robo-components`. We decided to keep RoboFile minimal.

RoboFile.php
```diff
@@ -71,4 +84,186 @@ public function __construct() {
...
+      }
+      elseif (strpos($settings_content, $cache_bin) !== FALSE) {
+        $this->say('Skipping: ' . $label . ' (already disabled)');
+      }
+    }
+
+    // Clear Drupal cache.
+    $this->taskExec('drush cr')->run();
+
+    $this->say("Drupal caching disabled.");
```
AronNovak, line 266: We do some file manipulations only, I'd try to verify it programmatically if the changes are truly effective and this way, we could do some error handling. Like doing some HTTP requests and check the headers or try to retrieve the config bits via the API, then we could have a conditional branch where we explain what went wrong and set a non-zero exit code. Same for the other direction.

mgurjanov:
@amitaibu In order to address PR review suggestions I'd need a TB bump.

amitaibu:
Sure, Bump to 5h

PR #707, opened 2024-09-11
Uninstall automated_cron, replaced by ultimate_cron

mariano-dagostino:


amitaibu:
What would be the use case of ultimate_cron ? (Or automated_cron).

amitaibu:
I mean both. Maybe we can remove both?

amitaibu:
> I see value in having ultimate_cron enabled

How would we typically use it? Or rather, what would be the rule of thumb, of when and how to configure? Maybe it's something we'd want to add to the README

AronNovak (approved):
Looks good to me. Automated Cron for me is not an equivalent for Ultimate Cron. Ultimate Cron is needed for fine-grained scheduling of jobs and fine-grained logging / error handling.
Automated Cron is needed for cheap hosting when nothing can invoke `drush core-cron` repeatedly.

amitaibu (approved):

Issue #714, opened 2024-09-13
DDEV Allow configuration to fail when a hook execution fails


mgurjanov:
If a ddev restart fails to run config import or similar build can pass as green because ddev fails to return non-zero exit code.

Config option added in https://github.com/ddev/ddev/issues/2536 and specifically code MR https://github.com/ddev/ddev/commit/5d17f0e5dccc321104489bc1ca9586186b6de552 allows us to fix this situation.

PR #715, opened 2024-09-13
DDEV Allow configuration to fail when a hook execution fails

mgurjanov:
#714 

- [x] fix ddev restart fail exit code


AronNovak (approved):
:100:

Issue #696, opened 2024-08-01
Fix Seckit connect-src form field limit [2h]


mgurjanov:
### Issue

Module `Seckit` has a character count limit on `connect-src` field that prevents us from adding large amount of entries. Service `google ads`  is using google domains per country which all would need to be added in `connect-src` form field. This is currently impossible - check [link](https://www.drupal.org/project/seckit/issues/3206728#comment-15616318).

### Todo 

Make a patch/MR for Seckit to make a change where it's possible to add more entries (up to a web servers response headers limit) and include in DS.

PR #717, opened 2024-09-15
Fix Seckit connect-src form field limit

mgurjanov:
#696 

- [x] require seckit ^2.0.3
- [x] patch text area to show more than 1 row by default for better UX

Note: Implemented per https://www.drupal.org/project/seckit/issues/3206728#comment-15760264

ds_seckit


amitaibu (approved):
:+1:

Issue #720, opened 2024-09-16
Configure text formats form hangs [3h]


mgurjanov:
Investigate if too many entries under `allowed_tags` option in https://github.com/Gizra/drupal-starter/blob/29ac99300d5199035bf2fe960de389dbf43be97e/config/sync/editor.editor.full_html.yml causes slowness in form opening/saving.

```
<a href hreflang> <em> <strong> <cite> <blockquote cite> <code> <ul type> <ol start type> <li> <dl> <dt> <dd> <h2 id> <h3 id> <h4 id> <h5 id> <h6 id> <s> <sup> <sub> <table> <caption> <tbody> <thead> <tfoot> <th> <td> <tr> <hr> <p> <h1> <pre> <drupal-media data-entity-type data-entity-uuid data-view-mode data-align data-caption alt title>
```

This is what umami_demo install for full html as allowed tags

PR #721, opened 2024-09-16
Configure text formats form hangs

mgurjanov:
#720 

- [x] changes to allowed tags for full html text format

Note: The issue did come from too large allowed tags string that needs to be parsed by CKEditor and compared to already implemented tags by filters or plugins. I added tags to Full HTML text format that are not supported by already configured filters or plugins. Now text formats form response is quick and there are no any errors.


mgurjanov:
@mariano-dagostino Very good point and I think allowing `<form>` in WYSIWYG can pose a security issue anyway.

amitaibu:
@mgurjanov Can you please explain the fix, I'm not sure I understand how it's working.

mgurjanov:
@amitaibu `Source editing` in text format configuration form is used to manually set `allowed tags as when we use filter `Limit allowed HTML tags and correct faulty HTML` we cannot anymore manually edit `Allowed HTML tags' since core 10.2. Since that update allowed tags are defined by enabled plugins and filters. So now when we add some tags to `Source editing` section, all those entries we add, on form save get parsed and compared against enabled plugins and filters. If for example `Bold` is enabled with plugin then one cannot add `<strong>` to this section again as it's already enabled. But we can add e.g. `<div class>` so div with class is allowed etc. We shouldn't add too much of these entries anyway as any developer who builds their site should define what extra attributes should be allowed in WYSIWYG, in `Full HTML` text format or any other text format. I am not sure but I think core comes without anything in this text area.
So I removed previous entries that were too many (you can check 2 yml files in the PR) and replaced it with some that could be added as they aren't already added by enabled plugins/filters.
Also I removed `<form>` as I think it poses a security risk. Usually forms are added to site in a different way not through WYSIWYG editor.
BTW this list of filters gets read in other places in code, not just when saving this form, so too many entries affects speed of site.

Source editing where change was made|Limit allowed HTML filter
-|-
allowed_tags|html_filter

mgurjanov:
@bboro Yes it has less but keep in mind that what I added in config can be seen in one yml, and the other yml holds my tags combined with other that enabled plugins/filters are adding.

@mariano-dagostino @bboro I also agree that we should probably not configure Full HTML by default.

Issue #718, opened 2024-09-16
Replace Twitter icon [1h]


isapisa:
### Issue

We should replace the twitter icon with the X icon: https://fontawesome.com/icons/x-twitter?f=brands&s=solid

https://gizra:1234@dev-gizra-drupal-starter.pantheonsite.io/news/current-digital-marketing-sports-watching-rather-marketing

<img width="1374" alt="Screenshot 2024-09-16 at 14 50 23" src="https://github.com/user-attachments/assets/73c99b73-f6cb-476a-893c-33d603f8ab8a">

### To Do

- Replace Twitter icon

balagan73:
@isapisa Should I also modify the hover text to Share on X?
![image](https://github.com/user-attachments/assets/8b60528e-330b-49ca-87fc-776027f90629)

isapisa:
@balagan73 yes please

sonvir249:
@amitaibu this seems to be fixed. Can we close this issue?
Current twitter icon:
![image](https://github.com/user-attachments/assets/8fe4ad33-2677-4f9e-9e4e-c2f97b70a2cf)

PR #724, opened 2024-09-19
Replaced twitter icon

balagan73:
#718 

- Replaced twitter icon, and adjusted viewbox.

TB: 0.6/1h


balagan73:
image

web/themes/custom/server_theme/templates/server-theme-social-share-button.html.twig
```diff
@@ -9,7 +9,7 @@
   {% set title = 'Share on LinkedIn'|t %}
 {% elseif service == 'twitter' %}
   {% set shareUrl = "https://twitter.com/intent/tweet?url=" ~ url.toString() %}
```
amitaibu, line 11: I think URL should change? Also, can we rename the `service` to `x`?

PR #729, opened 2024-10-07
Increase database log retention

AronNovak:
When there's an anomaly on a client site, we'd like to be able to trace back DBlog messages. In some cases, we already increased it. With server-side Rollbar problems we have, this is even more important, I think.

### DIFF
--- a/config/sync/dblog.settings.yml
+++ b/config/sync/dblog.settings.yml
@@ -1,3 +1,3 @@
-row_limit: 1000
+row_limit: 50000
 _core:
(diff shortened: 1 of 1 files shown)


config/sync/dblog.settings.yml
```diff
@@ -1,3 +1,3 @@
-row_limit: 1000
+row_limit: 50000
```
AronNovak, line 1:
```suggestion
row_limit: 100000
```

AronNovak:
@mariano-dagostino Thanks, indeed we should not ruin the functionality of the form.

PR #731, opened 2024-10-10
Recent News views page shows all translations

mgurjanov:
While working on a https://ns2024.drupal.rs/en we realized that News views page isn't filtering content per current page language but shows all translations.

- [x] filter recent News per current user translation language

ds_news

### DIFF
--- a/config/sync/views.view.news.yml
+++ b/config/sync/views.view.news.yml
@@ -201,6 +201,48 @@ display:
           expose:
             operator_limit_selection: false
             operator_list: {  }
+        langcode:
+          id: langcode
+          table: node_field_data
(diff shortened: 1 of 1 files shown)


bboro (approved):

amitaibu:
Oh, that site looks really nice! Hopefully by the end of the camp, more Serbians will know about Drupal-starter :smile:

amitaibu:
@mgurjanov can you please check if the Events view also requires a same fix? Otherwise, MWG

mgurjanov:
@amitaibu Thanks! I checked other views and this was the only requiring this change.

PR #733, opened 2024-10-11
Crowdsec

AronNovak:
https://www.drupal.org/project/crowdsec

> CrowdSec collects malicious IP addresses globally from everybody who is willing to collaborate. There are many [supported "scenarios"](https://hub.crowdsec.net/browse%22), under which bad actors and their IP addresses can be reported. Once CrowdSec is confident because the signals about an IP address are strong enough, that IP gets added to their public block list for a period of time.
> 
> This module integrates with the CrowdSec API using their PHP SDK to verify the IP address of each request getting to your Drupal site against their block list and should an IP be listed for being banned, those requests will be rejected and responded with a 403 HTTP response status code.
> 
> No worries, the block list is cached locally, and no performance implications are expected.

For several projects, we picked IPs by hand, as no WAF is available.
This is a collective global IP picking automated, round the clock :lock: !


AronNovak:
https://www.drupal.org/project/crowdsec/issues/3480150

AronNovak:
@mariano-dagostino Good point!
```php
  /** 
   * Initialize and return the cache backend.
   *
   * @return \CrowdSec\RemediationEngine\CacheStorage\AbstractCache
   *   The initialized cache backend.
   */
  private function cache(): AbstractCache {
    if (!isset($this->cache)) {
      if ($redis = Settings::get('redis.connection')) {
        $dns = 'redis://' . ($redis['host'] ?? 'localhost') . ':' . ($redis['port'] ?? '6379');
        $this->cache = new Redis(['redis_dsn' => $dns], $this->logger);
      }   
      else {
        $this->cache = new PhpFiles(['fs_cache_path' => $this->fileSystem->getTempDirectory() . '/crowdsec'], $this->logger);
      }   

    }   
    return $this->cache;
  }
```

This is from the code of the module. If Redis is configured, it automatically uses that, the doc is misleading in my opinion.

AronNovak:
It's serving one of the client sites for a while now. We may consider, at another site, @balagan73 faced with very strange Rollbar errors, if we'd have Crowdsec, we'd have less noise (and more power for visitors)

patches/crowdsec-cli.patch
```diff
@@ -0,0 +1,13 @@
+diff --git a/src/Middleware.php b/src/Middleware.php
```
amitaibu, line 1: Is this file needed?

amitaibu (commented):
Thanks. Let's add a word in the README

patches/crowdsec-cli.patch
AronNovak: Nope, the maintainer added it super quickly to the project.

AronNovak:
Added README as well.

README.md
```diff
@@ -519,3 +519,10 @@ when a privileged user will login, they must enable their TFA. Otherwise, on a s
 login, they will already be blocked. A site admin may reset their validation tries
 under the `/admin/people` page.
 The TFA method that is enabled is one that uses Google authenticator (or similar).
+
+## WAF - Crowdsec
+
+It is recommended to use a proper WAF, either from Cloudflare, or from another vendor, but
+for smaller sites, it is not always possible. We integrated [Crowdsec](https://www.crowdsec.net/)
```
amitaibu, line 526:
```suggestion
for smaller sites, it is not always possible.  [Crowdsec](https://www.crowdsec.net/) is integrated
```

To confirm, the recommendation is to have it always turned on, even with a WAF service?

AronNovak: @amitaibu Yep, it could proactively react to probing, that a WAF cannot do, at least the ones that we worked with.

amitaibu (approved):
Thanks

PR #748, opened 2024-11-12
Fixes sidebar button disappears when clicking it

### LINKED ISSUE #725: Create button on the sidebar disappears when clicking it
I guess the `aria-controls="admin-toolbar"` attribute was removed from the button, and now when clicking the Create button it will be hidden.

Before clicking|After clicking
-|-
image|image

Need to adjust web/themes/custom/server_theme/src/js/toggler.js

sonvir249:
Fixes #725


web/themes/custom/server_theme/src/js/toggler.js
```diff
@@ -14,7 +14,7 @@
    * @type {{attach: Drupal.behaviors.serverThemeToggler.attach}}
    */
   Drupal.behaviors.serverThemeToggler = {
-    attach: function(context, settings) {
+    attach: function (context, settings) {
       const $buttons = $(once('server-theme-button', 'button[aria-controls]', context)).not('button[aria-controls="admin-toolbar"]');
```
sonvir249, line 18: `aria-controls="admin-toolbar"` is not added by drupal core. Please check https://git.drupalcode.org/project/drupal/-/blob/10.3.x/core/modules/navigation/templates/navigation-menu.html.twig#L50

web/themes/custom/server_theme/src/js/toggler.js
```diff
@@ -26,6 +26,11 @@
           return;
         }
 
+        // Do nothing if it's a sidebar navigation toolbar button.
+        if ($this.hasClass('toolbar-button')) {
```
amitaibu, line 30: I think we can move that to the jQuery selector, from above (`const $buttons = $(once('server-theme-...`)  - so it won't include these kind of buttons.

sonvir249: @amitaibu Done, updated the PR. Can you please review it again?

web/themes/custom/server_theme/src/js/toggler.js
```diff
@@ -14,8 +14,8 @@
    * @type {{attach: Drupal.behaviors.serverThemeToggler.attach}}
    */
   Drupal.behaviors.serverThemeToggler = {
-    attach: function(context, settings) {
-      const $buttons = $(once('server-theme-button', 'button[aria-controls]', context)).not('button[aria-controls="admin-toolbar"]');
+    attach: function (context, settings) {
+      const $buttons = $(once('server-theme-button', 'button[aria-controls], button:not(.toolbar-button)', context)).not('button[aria-controls="admin-toolbar"]');
```
amitaibu, line 18: Untested, but maybe we can avoid having two `not`

```suggestion
        const $buttons = $(once('server-theme-button', 'button[aria-controls]:not([aria-controls="admin-toolbar"]), button:not(.toolbar-button)', context));
```

sonvir249: @amitaibu Indeed, that's a good Idea. Thanks for the suggestion. Updated the PR, Please check it.

#### Buttons before code change:
image

#### Buttons after code change:
image

amitaibu: I think the code is still calling the `not` method `$(...).not(...)`.

Can it be changed to `const $buttons = $(once('server-theme-button', 'button[aria-controls]:not([aria-controls="admin-toolbar"]), button:not(.toolbar-button)', context));`?

sonvir249: @amitaibu Did you get a chance to look at this?

amitaibu: Yes, I still think we can avoid the `.not()` function, and have it inside the once :)

web/themes/custom/server_theme/src/js/toggler.js
```diff
@@ -14,8 +14,9 @@
    * @type {{attach: Drupal.behaviors.serverThemeToggler.attach}}
    */
   Drupal.behaviors.serverThemeToggler = {
-    attach: function(context, settings) {
-      const $buttons = $(once('server-theme-button', 'button[aria-controls]', context)).not('button[aria-controls="admin-toolbar"]');
+    attach: function (context, settings) {
+      const $buttons = $(once('server-theme-button', 'button[aria-controls]:not(.toolbar-button)', context)).not('button[aria-controls="admin-toolbar"]');
```
amitaibu, line 18: I think we can have something like this:

```js
const $buttons = $(
  once(
    'server-theme-button', 
    'button[aria-controls]:not(.toolbar-button):not([aria-controls="admin-toolbar"])', 
    context
  )
);
```

sonvir249: @amitaibu it worked, thanks for the suggestion.

amitaibu (approved):
Thanks

PR #752, opened 2024-11-13
Rollbar integration

AronNovak:


AronNovak:
This is the 1-1 port from client project. We can improve it as agreed in a follow-up.

.travis.yml
```diff
@@ -12,6 +12,7 @@ stages:
 env:
   global:
   - DDEV_NO_INSTRUMENTATION=true
+  - ROLLBAR_SERVER_TOKEN=df6ce617465b4980afdecc95ed1b42de
   - PANTHEON_GIT_URL="ssh://codeserver.dev.5acc1750-abf0-47ef-a6d8-6890a6efaca8@codeserver.dev.5acc1750-abf0-47ef-a6d8-6890a6efaca8.drush.in:2222/~/repository.git"
```
amitaibu, line 16: Off topic - now with the Codespace integration, we can probably remove the Pantheon.

.travis.yml
```diff
@@ -12,6 +12,7 @@ stages:
 env:
   global:
   - DDEV_NO_INSTRUMENTATION=true
+  - ROLLBAR_SERVER_TOKEN=df6ce617465b4980afdecc95ed1b42de
```
amitaibu, line 15: Can we encrypt the token (and change this one, so it's not exposed)?

composer.patches.json
```diff
@@ -32,6 +32,9 @@
     },
     "drupal/seckit": {
       "d.o. #3206728": "https://www.drupal.org/files/issues/2024-09-06/seckit-2_x-src-field-length-3206728-8.patch"
+    },
+    "drupal/rollbar": {
+      "memory issues": "patches/3432364-rollbar-vars.patch"
```
amitaibu, line 37: Is there a do issue?

amitaibu (commented):
Thanks, a few comments

composer.patches.json
AronNovak: https://www.drupal.org/node/3432364

.travis.yml
AronNovak: This is not a real token, a dummy one from the documentation. It is enough if we try to send an item to Rollbar, if it fails or not, for the sake of this test, it does not matter.

AronNovak:
```
 aaron  ~  curl -H 'X-Rollbar-Access-Token: df6ce617465b4980afdecc95ed1b42de' 'https://api.rollbar.com/api/1/projects'
{
  "err": 1,
  "message": "invalid access token"
} aaron  ~  

```

.travis.yml
AronNovak: I guess it's more a PM question, if they need persistent demo stuff or not.
@isapisa @nora-novak

amitaibu: I don't think it's needed.

.travis.yml
amitaibu: Thanks, let's add a comment above it's dummy

composer.patches.json
amitaibu, line 37:
```suggestion
      "d.o. #3432364": "patches/3432364-rollbar-vars.patch"
```

amitaibu:
Needs a re-roll

amitaibu:
@AronNovak sorry, I merged a PR before, and it caused another conflict 😬

amitaibu (approved):

PR #754, opened 2024-11-14
Issue-693: Disallow scheduler on locked pages (if scheduler module is enabled) 

### LINKED ISSUE #693: Disallow scheduler on locked pages (if scheduler module is enabled)
### Issue

Locked pages can be unpublished by setting scheduler on a locked page.

### Todo

We should check on locked pages if scheduler is set and show a form error thus preventing locked page to become potentially unpublished.

sonvir249:
Closes #693


web/modules/custom/server_general/server_general.module
```diff
@@ -350,3 +376,21 @@ function server_general_preprocess_media_oembed_iframe(array &$variables) {
     $variables['media'] = IFrameMarkup::create(str_replace('youtube.com/', 'youtube-nocookie.com/', $variables['media']));
   }
 }
+
+/**
+ * Implements hook_scheduler_unpublishing_allowed().
+ */
+function server_general_scheduler_unpublishing_allowed(EntityInterface $entity) {
+  $allowed = TRUE;
+  if ($entity->getEntityType()->id() !== 'node') {
```
mgurjanov, line 385: I think we implemented this locked pages only for `Landing page` bundle, so we could maybe narrow down this check to `$entity->bundle() !== 'landing_page'` not to trigger this for every node.

web/modules/custom/server_general/server_general.module
```diff
@@ -350,3 +376,21 @@ function server_general_preprocess_media_oembed_iframe(array &$variables) {
     $variables['media'] = IFrameMarkup::create(str_replace('youtube.com/', 'youtube-nocookie.com/', $variables['media']));
   }
 }
+
+/**
+ * Implements hook_scheduler_unpublishing_allowed().
+ */
+function server_general_scheduler_unpublishing_allowed(EntityInterface $entity) {
+  $allowed = TRUE;
```
mgurjanov, line 384: We shouldn't set $allowed = TRUE by default as other modules could also call this hook. Instead I would just return FALSE where our locked page conditions are met and otherwise just return leaving our hook neutral.

web/sites/default/default.services.yml
```diff
@@ -2,7 +2,7 @@ parameters:
   # Toggles the super user access policy. If your website has at least one user
   # with the Administrator role, it is advised to set this to false. This allows
   # you to make user 1 a regular user, strengthening the security of your site.
-  security.enable_super_user: false
+  security.enable_super_user: true
```
mgurjanov, line 5: This has no connection to this PR.

web/modules/custom/server_general/server_general.module
```diff
@@ -225,6 +224,33 @@ function server_general_form_alter(array &$form, FormStateInterface $form_state)
...
+      foreach (array_keys($form) as $key) {
+        if (in_array($key, $scheduler_fields)) {
+          unset($form[$key]);
+        }
+      }
+    }
+    $current_time = time();
+    // If node is locked and schedule dates are set, make them readonly and disabled.
+    if (($publish_on > $current_time) && ($unpublish_on > $current_time) && $locked_pages_service->isNodeLocked($entity)) {
+      $form['scheduler_settings']['#description'] = t('This is the help text for the details section.');
```
mgurjanov, line 246: This text could explain why these options are read only.

mgurjanov (changes requested):
Overall nice attempt at this issue! I left some change requests, though.

web/modules/custom/server_general/server_general.module
sonvir249: @mgurjanov this is done.

web/modules/custom/server_general/server_general.module
sonvir249: @mgurjanov this is done.

web/sites/default/default.services.yml
sonvir249: @mgurjanov removed this file.

web/modules/custom/server_general/server_general.module
sonvir249: Displaying the description as an error meesage. Please do let me know if any changes needed. Thanks!
image

web/modules/custom/server_general/server_general.module
```diff
@@ -225,6 +224,39 @@ function server_general_form_alter(array &$form, FormStateInterface $form_state)
...
+      }
+    }
+    $current_time = time();
+    // If node is locked and schedule dates are set, make them readonly and disabled.
+    if (($publish_on > $current_time) && ($unpublish_on > $current_time) && $locked_pages_service->isNodeLocked($entity)) {
+      $form['scheduler_settings']['#description'] = [
+        '#type' => 'inline_template',
+        '#template' => '<strong class="action-link--danger">{{ message }}</strong>',
+        '#context' => [
+          'message' => t('This page is locked hence the publish and unpublish dates can not be changed. Also schedling dates won\'t work.'),
```
mgurjanov, line 250:
```suggestion
          'message' => t('This page is locked hence the publish and unpublish dates can not be changed. Also scheduling dates won\'t work.'),
```

mgurjanov (approved):
Looks good to me. Just one small typo fix.

sonvir249:
@amitaibu @mgurjanov Just a suggestion, shouldn't we add github actions to catch  the drupal php and twig code stanrdard, theme build issues whenever any PR is rasied or commit is pushed.

mgurjanov:
@sonvir249 We do run `ddev phpcs` or `ddev phpstan` locally prior to pushing code to github.

web/modules/custom/server_general/server_general.module
sonvir249: @mgurjanov fixed the spelling mistake.

sonvir249:
> @sonvir249 We do run `ddev phpcs` or `ddev phpstan` locally prior to pushing code to github.

I also do that on local, but having it in the pipeline adds an extra layer of assurance to ensure nothing goes wrong with the code.

web/modules/custom/server_general/server_general.module
```diff
@@ -225,6 +224,39 @@ function server_general_form_alter(array &$form, FormStateInterface $form_state)
   /** @var \Drupal\server_general\LockedPages $locked_pages_service */
   $locked_pages_service = \Drupal::service('server_general.locked_pages');
 
+  /** @var \Drupal\Core\Extension\ModuleHandlerInterface $module_handler */
+  $module_handler = \Drupal::service('module_handler');
```
amitaibu, line 228: Let's move this to a helper function.
Also, let's use the style if return early. So

```
$module_handler = \Drupal::service('module_handler');
if (!$module_handler->moduleExists('scheduler')) {
  // Scheduler module not enabled.
  return;
}

File: web/modules/custom/server_general/server_general.module [hunk 7]
```diff
@@ -225,6 +224,39 @@ function server_general_form_alter(array &$form, FormStateInterface $form_state)
...
+    // If node is locked and schedule dates are not set, remove publish_on, unpublish_on and scheduler_settings.
+    if (!$publish_on && !$unpublish_on && $locked_pages_service->isNodeLocked($entity)) {
+      $scheduler_fields = ['publish_on', 'unpublish_on', 'scheduler_settings'];
+      foreach (array_keys($form) as $key) {
+        if (in_array($key, $scheduler_fields)) {
+          unset($form[$key]);
+        }
+      }
+    }
+    $current_time = time();
```
amitaibu, line 243: https://api.drupal.org/api/drupal/core%21lib%21Drupal%21Component%21Datetime%21TimeInterface.php/function/TimeInterface%3A%3AgetCurrentTime/9#:~:text=Returns%20the%20current%20system%20time,time%20the%20method%20was%20called.&text=%24request_time%20%3D%20%5CDrupal%3A%3Atime()%2D%3EgetCurrentTime()%3B

mgurjanov:
@amitaibu I was thinking of asking for test case, but wouldn't that require installing and enabling scheduler?

web/modules/custom/server_general/server_general.module
```diff
@@ -225,6 +224,39 @@ function server_general_form_alter(array &$form, FormStateInterface $form_state)
...
+    // If node is locked and schedule dates are set, make them readonly and disabled.
+    if (($publish_on > $current_time) && ($unpublish_on > $current_time) && $locked_pages_service->isNodeLocked($entity)) {
+      $form['scheduler_settings']['#description'] = [
```
amitaibu, line 246: Please share a screenshot. Do we have something similar on Drupal-starter?

web/modules/custom/server_general/server_general.module
```diff
@@ -350,3 +382,20 @@ function server_general_preprocess_media_oembed_iframe(array &$variables) {
     $variables['media'] = IFrameMarkup::create(str_replace('youtube.com/', 'youtube-nocookie.com/', $variables['media']));
   }
 }
+
+/**
+ * Implements hook_scheduler_unpublishing_allowed().
+ */
+function server_general_scheduler_unpublishing_allowed(EntityInterface $entity) {
+  /** @var \Drupal\node\NodeInterface $entity */
```
amitaibu, line 390: I don't think the entity is always `NodeInterface`. In fact, we already have type hinting above `EntityInterface $entity`

amitaibu:
> I was thinking of asking for test case,

@sonvir249 always go with that gut feeling 😄 

> but wouldn't that require installing and enabling scheduler?

We can have it installed with composer, and the test would enable it. In the future, we could decide to also enable it by default - depends if we see more real world cases where clients need it.

mgurjanov:
@amitaibu It's funny how it always fires back to me when I don't, and I do follow gut feeling most of the time. Thanks for the reminder!

amitaibu:
😆 👌

web/modules/custom/server_general/server_general.module
sonvir249: @amitaibu This is done.

web/modules/custom/server_general/server_general.module
sonvir249: @amitaibu this is done.

web/modules/custom/server_general/server_general.module
sonvir249: @amitaibu this is done.

sonvir249:
> > ddev phpcs or ddev phpstan
> 
> They are already part of the [Travis](https://github.com/Gizra/drupal-starter/blob/026a54717eb699f8c47d2034940a65da1dd24746/.travis.yml) CI pipieline
> 
> @sonvir249 could you also add a test case please?

@amitaibu Sure, I can handle that. Could you help clarify what areas I should cover in the tests? Also, I don’t have much experience writing test cases, so any references or guidance would be really helpful. Thanks!

amitaibu:
You can confirm, for example, that the form changes have kicked in.

You can see an existing test [here](https://github.com/Gizra/drupal-starter/blob/09fd86c18565367a819ecea5aa3a5aa4ddb7af70/web/modules/custom/server_general/tests/src/ExistingSite/ServerGeneralNodeLandingPageTest.php)

sonvir249:
> You can confirm, for example, that the form changes have kicked in.
> 
> You can see an existing test [here](https://github.com/Gizra/drupal-starter/blob/09fd86c18565367a819ecea5aa3a5aa4ddb7af70/web/modules/custom/server_general/tests/src/ExistingSite/ServerGeneralNodeLandingPageTest.php)

@amitaibu I have added the basic unit test to verify the scheduler form fields on locked and non-locked node forms.

sonvir249:
@amitaibu @mgurjanov I faced https://www.drupal.org/project/config_pages/issues/3474241 deprecation notice while removing configured node. Should we apply this patch for `config_pages` module?

config/sync/core.extension.yml
```diff
@@ -71,6 +71,7 @@ module:
   redirect: 0
   redis: 0
   responsive_image: 0
+  scheduler: 0
```
amitaibu, line 74: As this is not a very common feature, I don't think we should enable this by default. It's fine that we download `scheduler`, but we should enable it only from the test.

sonvir249: @amitaibu this is done. Enabling and configuring `scheduler` module from the test.

web/sites/default/default.services.yml
amitaibu: This shouldn't be changed

web/modules/custom/server_general/server_general.module
```diff
@@ -353,3 +387,20 @@ function server_general_preprocess_media_oembed_iframe(array &$variables) {
...
+  if ($entity->bundle() !== 'landing_page') {
+    return;
+  }
+
+  /** @var \Drupal\server_general\LockedPages $locked_pages_service */
+  $locked_pages_service = \Drupal::service('server_general.locked_pages');
+
+  if ($locked_pages_service->isNodeLocked($entity)) {
+    return FALSE;
+  }
```
amitaibu, line 405:
```suggestion
  return !$locked_pages_service->isNodeLocked($entity)
```

web/modules/custom/server_general/src/LockedPages.php
```diff
@@ -30,17 +31,27 @@ class LockedPages {
    */
   protected $entityFieldManager;
 
+  /**
```
amitaibu, line 34: Changes on this file can be removed

web/modules/custom/server_general/src/LockedPages.php
```diff
@@ -109,4 +120,17 @@ public function getReferencedBundles(): array {
...
+   * Checks if module is enabled.
+   *
+   * @param string $module_name
+   *   The module name.
+   *
+   * @return bool
+   *   Returns TRUE if module is enabled else FALSE.
+   */
+  public function isModuleEnabled(string $module_name): bool {
```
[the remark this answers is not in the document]
sonvir249: @amitaibu this is done.

web/modules/custom/server_general/src/LockedPages.php
sonvir249: @amitaibu this is done.

web/modules/custom/server_general/server_general.module
```diff
@@ -228,6 +227,41 @@ function server_general_form_alter(array &$form, FormStateInterface $form_state)
...
+  $request_time = \Drupal::time()->getCurrentTime();
+  // If node is locked and schedule dates are set,
+  // make them readonly and disabled.
+  if (($publish_on > $request_time) && ($unpublish_on > $request_time) && $locked_pages_service->isNodeLocked($entity)) {
+    $form['scheduler_settings']['#description'] = [
+      '#type' => 'inline_template',
+      '#template' => '<strong class="action-link--danger">{{ message }}</strong>',
+      '#context' => [
+        'message' => t("This page is locked hence the publish and unpublish dates can not be changed. Also scheduling dates won't work."),
+      ],
```
amitaibu, line 257: Let's create a twig, and move it there, so we don't end up with css classes inside PHP.

We can have a `wrapDanger` function, like our other [wrappers](https://github.com/Gizra/drupal-starter/blob/73d4d58d45dfb3a21f4235b615640f8e560b2944/web/modules/custom/server_general/src/ElementWrapTrait.php)

sonvir249: @amitaibu Since we are modifying the node edit form, which utilizes the admin theme, any changes made in the `server_theme` theme will not apply to the admin theme.

amitaibu: Right, thanks

amitaibu (approved):
Thanks 🙏

Issue #726, opened 2024-09-25
Theming the default contact form [3h]


isapisa:
### Issue

`/contact`

<img width="1908" alt="Screenshot 2024-09-25 at 10 39 22" src="https://github.com/user-attachments/assets/39f851cd-cb71-4840-9c8e-c5222d628de3">

amitaibu:
Yes, thanks

sonvir249:
@amitaibu is this still applicable?
I can see the themed contact form:
![image](https://github.com/user-attachments/assets/0b6cad32-6bea-4896-bf0b-cd70bcc2de5c)

amitaibu:
Yes, it does. 

1. The label and value have weird gaps
2. Emali address is probaly a link, but it doesn't look like it.
3. Checkbox should be next to "Send yourself a copy"

<img width="1003" alt="image" src="https://github.com/user-attachments/assets/5fc43284-163d-448c-98b1-a0c28ed820de">

sonvir249:
@amitaibu raised a PR please review it.
FYI: We should display the email as plain text only, since it's for display purpose. Making it a link would trigger the email app to open.

mgurjanov:
This is done.

PR #759, opened 2024-11-21
Add Webform paragraph, and theme the contact form

mgurjanov:
#726 

- [x] install and enable webform module
- [x] add Form PT with webform reference field
- [x] wire and style Form PT
- [x] enable Form PT on Landing Page CT
- [x] update Contact Webform to match core Contact form (submit confirmation email is set to be sent to user as ntification email to site owner)
- [x] add tests
- [x] uninstall core Contact module

Edit|View
-|-
ds_form_pt_edit|ds_form_pt


mgurjanov:
@mariano-dagostino You stole my question! I was thinking of same. So you agree?

mgurjanov:
@mariano-dagostino I removed core contact!

mgurjanov:
@sonvir249 Please take a look.

config/sync/user.role.anonymous.yml
```diff
@@ -17,6 +16,5 @@ weight: -10
 is_admin: false
 permissions:
   - 'access content'
-  - 'access site-wide contact form'
```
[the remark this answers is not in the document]
mgurjanov: Well isn't this connected to core Contact and we can setup webform permissions for each separate webform:

webform_access

web/themes/custom/server_theme/src/pcss/form.pcss
```diff
@@ -13,3 +13,35 @@ form {
...
+
+  .form-item {
+    @apply flex flex-col space-y-2;
+  }
+
+  .form-required:after {
+    @apply mx-0.5;
+    content: "*";
+    color: #dc2323;
+  }
```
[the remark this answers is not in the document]
mgurjanov: Code is based on another project and I think there was an issue where asterisk for required fields wasn't visible. This covers it to be safe.

web/modules/custom/server_general/src/FormsEmbedTrait.php
```diff
@@ -0,0 +1,80 @@
+<?php
+
+namespace Drupal\server_general;
+
+use Drupal\Core\Cache\CacheableMetadata;
+use Drupal\paragraphs\ParagraphInterface;
+
+/**
+ * Views Embed Trait.
```
bboro, line 9: Comment needs updating

web/modules/custom/server_general/src/FormsEmbedTrait.php
```diff
@@ -0,0 +1,80 @@
...
+   * @param string|null $field_description
+   *   Optional description field name.
+   *
+   * @return array
+   *   Render array.
+   *
+   * @throws \Drupal\Component\Plugin\Exception\InvalidPluginDefinitionException
+   * @throws \Drupal\Component\Plugin\Exception\PluginNotFoundException
+   */
+  protected function buildWebformEmbed(ParagraphInterface $entity, array $build, string $webform_field_name = 'field_webform', ?string $field_title = NULL, ?string $field_description = NULL): array {
```
bboro, line 37: Should we pass in the webform render array, the title & description to the trait instead of passing an entity and field names? That fits the existing convention better imo.

web/modules/custom/server_general/src/Plugin/EntityViewBuilder/ParagraphForm.php
```diff
@@ -0,0 +1,45 @@
...
+   * @param array $build
+   *   The existing build.
+   * @param \Drupal\paragraphs\ParagraphInterface $entity
+   *   The entity.
+   *
+   * @return array
+   *   Render array.
+   */
+  public function buildFull(array $build, ParagraphInterface $entity): array {
+    return $this->buildWebformEmbed($entity, $build, 'field_webform', 'field_title', 'field_description');
```
bboro, line 42:
```suggestion
    $build[] = $this->buildWebformEmbed($entity, $build, 'field_webform', 'field_title', 'field_description');
    return $build;
```

web/modules/custom/server_general/src/FormsEmbedTrait.php
mgurjanov: I noticed that and I will make changes per your request. I am curious to why was this decision made, because that doesn't seem to be flexible enough so that I can reuse smaller components. Can you explain this a bit? Thanks.

mgurjanov: I do get it for this example and it's a valid point where we extract data in PEVB and just pass to methods in traits.

amitaibu: PEVB should be the one responsible for extracting the data from the entities and passing the raw data to the theming traits.

One way to see it is that when we create a style guide, we can call the traits with hardcoded data, and it will Just Work ™️

web/modules/custom/server_general/src/Plugin/EntityViewBuilder/ParagraphForm.php
amitaibu: So here we should extract the data from the entity, and hand it over to the theming traits

web/themes/custom/server_theme/src/pcss/form.pcss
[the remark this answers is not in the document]
amitaibu: Let's confirm it's indeed needed. How does it look without it?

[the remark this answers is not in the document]
mgurjanov: I confirm that without this code we have no asterisks at all on all required fields.

web/modules/custom/server_general/src/Plugin/EntityViewBuilder/ParagraphForm.php
```diff
@@ -0,0 +1,80 @@
...
+    if (!$entity->get('field_description')->isEmpty()) {
+      $elements[] = $this->buildProcessedText($entity, 'field_description');
+    }
+
+    // Build the webform.
+    /** @var ?\Drupal\webform\WebformInterface $webform */
+    $webform = $this->entityTypeManager->getStorage('webform')->load($webform_name);
+    if (empty($webform) || !$webform->isOpen()) {
+      return [];
+    }
```
amitaibu, line 65: Move up the func

web/modules/custom/server_general/src/Plugin/EntityViewBuilder/ParagraphForm.php
```diff
@@ -0,0 +1,80 @@
...
+   *
+   * @param array $build
+   *   The existing build.
+   * @param \Drupal\paragraphs\ParagraphInterface $entity
+   *   The entity.
+   *
+   * @return array
+   *   Render array.
+   */
+  public function buildFull(array $build, ParagraphInterface $entity): array {
```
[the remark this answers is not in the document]
amitaibu: What we should improve here, is that PEVB shouldn't be calling stuff like `buildParagraphTitle` directly.

Instead it would call a trait with a `buildWebform(string $title, array $description = [], array $form = []): array();`

By having this `buildWebform`, we could show it on the style guide.

So it's not about calling traits from other traits. It's about separation of conecerns. PEVB is about extracting the dynamic data, and the FE traits are about getting raw data, and theming it.

Does it answer? :)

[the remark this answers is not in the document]
bboro: For me it's also about knowing where things come from. So we have `server-theme-accordion.html.twig`? Then we have `AccordionTrait`, `accordion.pcss`, `accordion.js` etc and that's super easy to navigate with Ctrl+N on phpstorm. Before when we have `InnerElementTrait` and `InnerElementLayoutTrait` etc it was super confusing to me at least and also tedious when trying to work out a general overview of how the element is put together.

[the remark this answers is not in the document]
mgurjanov: Thank you both. I get it now. I was a bit confused with some implementations on other projects where we have not so uniform implementation.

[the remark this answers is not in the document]
bboro: @mgurjanov no worries, and yes other projects may be confusing because we were still working out how to do this :)

For me personally, I'd prefer "easier to understand/maintain" more than "elegant". So for example, personally I'd prefer doing the layout, styling etc of a piece of text in its own `server-theme-element--` twig, rather than have it split over multiple smaller `server-theme-inner-element-layout--` etc twigs. 

But for much smaller elements like `button`, `responsive-font-size`, `vertical-spacing` etc, it makes sense to have traits that do this, so it's easier to maintain (change one twig and the whole site is updated)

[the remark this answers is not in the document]
amitaibu: > For me personally, I'd prefer "easier to understand/maintain" more than "elegant"

Same 😄 

> personally I'd prefer doing the layout, styling etc of a piece of text in its own server-theme-element-- twig, rather than have it split over multiple smaller server-theme-inner-element-layout-- etc twigs.

@bboro can you please give an example, I'm not sure I understand.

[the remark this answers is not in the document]
bboro: I mean I personally prefer having `server-theme-element--awesome-card.html.twig` which defines its own `grid grid-cols-2` etc layout, than having to compose this through `wrapInnerElementLayoutAwesomeCard`.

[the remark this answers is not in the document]
amitaibu: I see your point; thanks. But I'd still advocate being stingy with creating new Twigs, as I’ve seen it’s all too easy to add custom styles rather than reuse existing ones within the Twig.

So, as we break things down into smaller pieces, the goal is to guide devs away from adding new Twigs unnecessarily. New Twigs are the devil! 🤣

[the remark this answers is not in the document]
bboro: :rofl: Or maybe you can think of them as just an ever growing number of minions. :smile:  :smile:  :smile:  :smile:

web/modules/custom/server_general/src/Plugin/EntityViewBuilder/ParagraphForm.php
```diff
@@ -0,0 +1,59 @@
...
+      return [];
+    }
+
+    $title = $this->getTextFieldValue($entity, 'field_title');
+    $description = $this->buildProcessedText($entity, 'field_description');
+    $webform = $this->getWebform($webform_name);
+
+    if (empty($webform)) {
+      return [];
+    }
```
amitaibu, line 52: Move above

web/modules/custom/server_general/tests/src/Traits/ParagraphCreationTrait.php
```diff
@@ -0,0 +1,32 @@
...
+  /**
+   * Creates a Paragraph and marks it for automatic cleanup.
+   *
+   * @param array $settings
+   *   The settings to pass to Paragraph creation.
+   *
+   * @return \Drupal\paragraphs\ParagraphInterface
+   *   The Paragraph.
+   */
+  protected function createParagraph(array $settings = []): ParagraphInterface {
```
amitaibu, line 25: Is this needed? How did we do tests until now?

web/modules/custom/server_style_guide/src/Controller/StyleGuideController.php
```diff
@@ -892,4 +905,18 @@ protected function getExpandingText(?int $lines_to_clamp = NULL, ?string $button
     return $this->wrapContainerWide($this->buildElementExpandingText($element, $lines_to_clamp, $button_label_more, $button_label_less));
   }
 
+  /**
+   * Get Webform element.
+   *
+   * @return array
+   *   The render array.
+   */
+  protected function getWebformElement(): array {
```
amitaibu, line 914: 👌

amitaibu (commented):
Looks good. Let's add a default content, a `/contact` Landing page with the Contact form, and add it as a menu item to the header.

web/modules/custom/server_general/tests/src/Traits/ParagraphCreationTrait.php
mgurjanov: We were repeating code like:

```
    $cta = Paragraph::create(['type' => 'cta']);
    $cta->set('field_title', 'Lorem ipsum dolor sit amet');
    $cta->set('field_body', 'Lorem ipsum dolor sit amet, consectetur adipiscing elit.');
    $cta->set('field_link', [
      'uri' => 'https://example.com',
      'title' => 'Button text',
    ]);
    $cta->save();
    $this->markEntityForCleanup($cta);
```

Should I do the same?

amitaibu: Then let's open a follow up, to replace it with a helper, and remove it from this PR (So our code will be consistent)

mgurjanov: https://github.com/Gizra/drupal-starter/issues/766

mgurjanov:
@amitaibu Can this be merged now?

amitaibu (approved):
Nice addition, thank you 🙏

PR #763, opened 2024-11-25
Pantheon default config fixes for PHP and Solr

AronNovak:
We must explicitly specify Solr 8, otherwise we get Solr 4, and cryptic error messages about incompatibility.

### DIFF
--- a/web/sites/default/settings.pantheon.php
+++ b/web/sites/default/settings.pantheon.php
@@ -137,5 +137,5 @@ if (file_exists(__DIR__ . '/settings.fast404.php')) {
 }
 
 $config['search_api.index.server_dev']['server'] = 'pantheon_solr8';
-// This only exists at DDEV.
+// As we push to Solr config of DDEV to Pantheon as well, we disable it here.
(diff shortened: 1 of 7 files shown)


web/sites/default/settings.pantheon.php
```diff
@@ -135,3 +135,7 @@
 if (file_exists(__DIR__ . '/settings.fast404.php')) {
   include __DIR__ . '/settings.fast404.php';
 }
+
+$config['search_api.index.server_dev']['server'] = 'pantheon_solr8';
+// This only exists at DDEV.
```
amitaibu, line 140: Do we need this in this file?

AronNovak: The comment or the next line? If we don't do `$config['search_api.server.solr']['status'] = FALSE;` this, Search Api will throw warnings about the unavailable server. And it ruins out notifier experience. Just saw it at a client project, it's a must in some way.

amitaibu (approved):

PR #765, opened 2024-12-02
Upgrade `coder` and `sirbrillig/phpcs-variable-analysis`

AronNovak:


AronNovak:
This is not needed as-is, it was to reproduce an anomaly.

PR #769, opened 2024-12-10
Use vertical spacings as imports in other elements

bboro:


amitaibu (approved):
Thanks, I really like this change. It's another step of re-using elements 🙏

PR #771, opened 2024-12-18
Upgrade DDEV + typo fix

AronNovak:


AronNovak:
It was just to try to reproduce some failures, but alone this makes sense to make sure latest DDEV works still.

amitaibu (approved):
👍

PR #772, opened 2024-12-20
Update Tailwind to 3.4 to allow subgrid and update news card to use subgrid as example.

dipakmdhrm:
### Update:
- [x] Update Tailwind to 3.4 to allow subgrid
- [x] Update news card to use subgrid as example.

-|Before|After
-|-|-
Desktop|desktop-before|desktop-after
Tablet|tablet-before|tablet-after
Mobile|mobile-before|mobile-after

You'll notice for Desktop and Tablet that after using subgrid, each element of a card is in alignment with the corresponding element of all other cards in the same row, which wasn't the case before subgrid. Check description for eg.

### DIFF
--- a/web/modules/custom/server_general/src/ElementTrait/NewsTeasersTrait.php
+++ b/web/modules/custom/server_general/src/ElementTrait/NewsTeasersTrait.php
@@ -63,26 +63,32 @@ trait NewsTeasersTrait {
    *   Render array.
    */
   protected function buildElementNewsTeaser(array $image, string $title, Url $url, array $summary, int $timestamp): array {
-    $elements = [];
-
     // Labels.
-    $element = $this->buildLabelsFromText([$this->t('News')]);
-    $elements[] = $this->wrapTextResponsiveFontSize($element, 'sm');
+    $labels = $this->buildLabelsFromText([$this->t('News')]);
+    $labels = $this->wrapTextResponsiveFontSize($labels, 'sm');
 
     // Date.
-    $element = IntlDate::formatPattern($timestamp, 'short');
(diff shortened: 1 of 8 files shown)


dipakmdhrm:
More work is needed here. subgrid elements are inheriting the parent's gap which is messing with the gap between elements.

web/modules/custom/server_general/src/ElementTrait/NewsTeasersTrait.php
```diff
@@ -63,26 +63,32 @@ protected function buildElementNewsTeasers(string $title, array $body, array $it
...
+    $title = $this->wrapTextLineClamp($title, 3);
 
     // Body teaser.
-    $elements[] = $this->wrapTextLineClamp($summary, 4);
-
-    return $this->buildInnerElementLayoutWithImage($url, $image, $elements);
+    $description = $this->wrapTextLineClamp($summary, 4);
+
+    return [
+      '#theme' => 'server_theme_element__news_card',
```
amitaibu, line 84: We should not take this approach. Our approach is in fact the exact opposite. We prefer to break down those twigs, in favor of composing elements -- just like in the existing code.

web/themes/custom/server_theme/server_theme.theme
```diff
@@ -484,6 +484,17 @@ function server_theme_theme() {
...
+    'variables' => [
+      'image' => NULL,
+      'labels' => NULL,
+      'date' => NULL,
+      'title' => NULL,
+      'description' => NULL,
+      'url' => NULL,
+    ],
+  ];
+
```
amitaibu, line 497: ❌

Issue #766, opened 2024-12-04
Implement ParagraphCreationTrait for tests and replace where needed [2h]


mgurjanov:
Instead of creating paragraph in each test implement ParagraphCreationTrait like:

```
<?php

namespace Drupal\Tests\server_general\Traits;

use Drupal\paragraphs\Entity\Paragraph;
use Drupal\paragraphs\ParagraphInterface;
use weitzman\DrupalTestTraits\DrupalTrait;

/**
 * Wraps the node creation trait to track entities for deletion.
 */
trait ParagraphCreationTrait {

  use DrupalTrait;

  /**
   * Creates a Paragraph and marks it for automatic cleanup.
   *
   * @param array $settings
   *   The settings to pass to Paragraph creation.
   *
   * @return \Drupal\paragraphs\ParagraphInterface
   *   The Paragraph.
   */
  protected function createParagraph(array $settings = []): ParagraphInterface {
    $entity = Paragraph::create($settings);
    $entity->save();
    $this->markEntityForCleanup($entity);
    return $entity;
  }

}
```

PR #774, opened 2024-12-22
Implement ParagraphCreationTrait for tests and replace where needed

mgurjanov:
#766 

- [x] add paragraph creation trait
- [x] use method from paragraph creation trait in tests


amitaibu (approved):
Thanks

PR #781, opened 2025-01-09
Robust local `phpunit` test execution

AronNovak:
Since `crowdsec`, here we cannot execute `phpunit` twice, as we reach the number of 4xx attempts.

Also we have always failing Rollbar test, let's only execute that inside Travis due to the Rollbar key, which is a requirement for that.


patches/crowdsec_exclude_private_ips.patch
```diff
@@ -0,0 +1,52 @@
+diff --git a/src/EventSubscriber/CrowdSec.php b/src/EventSubscriber/CrowdSec.php
```
AronNovak, line 1: To get rid of this patch, we might do `$config['crowdsec.settings']['whisper']['enable'] = 0;`, explanation:
```
@aron novak you're right that pipelines should be identical to the live system, that's our philosophy as well. And yet, we're looking to behave differently. With the dry-mode, the question remains where you draw the line. Doing everything except the ban is almost exactly the same as disabling the whisper handling. You can see that in \Drupal\crowdsec\Buffer::addWhisperSignal where the actual ban happens 10 lines later than the decision to return because of the disabled whisper. All the lines in-between are pure variable collection. And it should be the responsibility of this module to make sure that those work correctly. The live site tests should not rely on them, do they?
```

.ddev/commands/web/phpunit
```diff
@@ -1,7 +1,15 @@
...
 ## Example: "ddev phpunit" or "ddev phpunit web/modules/custom/server_general".
 
-./vendor/bin/phpunit -c phpunit.xml.dist  $@
+if [ $# -eq 0 ]; then
+  # If no arguments are provided, run non-Rollbar tests
+  ./vendor/bin/phpunit -c phpunit.xml.dist --do-not-cache-result --exclude-group=Rollbar || (ddev drush watchdog-show --count=20 && exit 1)
+else
+  # If arguments are provided, pass them to PHPUnit
+  ./vendor/bin/phpunit -c phpunit.xml.dist "$@" || (ddev drush watchdog-show --count=20 && exit 1)
+fi
```
amitaibu, line 13: What commands are we sending (with and without Rollbar)?

web/sites/default/settings.ddev.php
```diff
@@ -105,3 +105,6 @@
 if (file_exists(__DIR__ . '/settings.fast404.php')) {
   include __DIR__ . '/settings.fast404.php';
 }
+
+// Let tests from PhpUnit do strange requests.
```
amitaibu, line 109:
```suggestion
//Disable CrowdSec's "whisper" locally. So one doesn't get blocked locally, or PHPUnit can work well.
```

.ddev/commands/web/phpunit
AronNovak: So the idea is if we do `ddev phpunit`, Rollbar test is excluded completely.
Travis will always execute it, it's handled in https://github.com/Gizra/drupal-starter/blob/main/ci-scripts/test_phpunit_rollbar.sh

It is also possible to execute Rollbar tests locally, if you specify anything extra, like:
`ddev phpunit --group=Rollbar`, it will execute it without issues.

Does this answer your question?

.ddev/commands/web/phpunit
```diff
@@ -1,7 +1,13 @@
...
 
 ## Description: Run PHPunit tests inside the web container.
-## Usage: phpunit
+## Usage: phpunit [arguments]
 ## Example: "ddev phpunit" or "ddev phpunit web/modules/custom/server_general".
 
-./vendor/bin/phpunit -c phpunit.xml.dist  $@
+if [ $# -eq 0 ]; then
+  # If no arguments are provided, run non-Rollbar tests
+  ./vendor/bin/phpunit -c phpunit.xml.dist --do-not-cache-result --exclude-group=Rollbar || (drush watchdog-show --count=20 && exit 1)
```
AronNovak, line 9:
```suggestion
  ./vendor/bin/phpunit -c phpunit.xml.dist --exclude-group=Rollbar || (drush watchdog-show --count=20 && exit 1)
```

.ddev/commands/web/phpunit
```diff
@@ -1,7 +1,13 @@
...
 
 ## Description: Run PHPunit tests inside the web container.
-## Usage: phpunit
+## Usage: phpunit [arguments]
 ## Example: "ddev phpunit" or "ddev phpunit web/modules/custom/server_general".
 
-./vendor/bin/phpunit -c phpunit.xml.dist  $@
+if [ $# -eq 0 ]; then
+  # If no arguments are provided, run non-Rollbar tests
+  ./vendor/bin/phpunit -c phpunit.xml.dist --exclude-group=Rollbar
```
amitaibu, line 9: Sorry, I still don't understand why we exclude Rollbar

AronNovak: Because it needs a Rollbar API key to be executed, which is not present on `local`. I'd not commit a key to the repository as-is, as it's antipattern and we could easily exceed rate limits.

amitaibu (approved):
:+1:

PR #784, opened 2025-01-11
Robustness fixes for the deploy process

AronNovak:
- better readability of Travis output by using `-x` - knowing which command resulted in the given error message by printing the command itself automatically
 - remove newline from the Travis commit message, it can easily break Travis config syntax
 - do not try to clone the repo if it already exists (when Travis tries to retry the failed deploy)


amitaibu (approved):
👍

PR #790, opened 2025-01-28
Focal point doesn't work properly when the preview image for the focal_point widget uses an image style that uses a focal_point image effect

dipakmdhrm:
Reference: https://www.drupal.org/node/2872960

### Update:
- Added a new 'Focal point preview' image style with simple '480px width scale' effect.
- Update the Image field focal_point widget in Image media entity to use this image style.

Before|After
-|-
https://github.com/user-attachments/assets/689fa5b9-29f4-4039-8adb-44d73e69b3b6|https://github.com/user-attachments/assets/f6e09d3e-3504-488c-84d1-924ace42257a

### DIFF
--- a/config/sync/core.entity_form_display.media.image.default.yml
+++ b/config/sync/core.entity_form_display.media.image.default.yml
@@ -6,7 +6,7 @@ dependencies:
     - field.field.media.image.field_caption
     - field.field.media.image.field_media_credit
     - field.field.media.image.field_media_image
-    - image.style.large
+    - image.style.focal_point_preview
     - media.type.image
   module:
     - focal_point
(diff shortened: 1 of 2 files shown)


AronNovak (approved):
Looks good. Do you happen to have a screenshot of the result? Is it ported from a client project? Does it affect all of our projects with Focal Point module installed? :thinking:

dipakmdhrm:
> Looks good. Do you happen to have a screenshot of the result?
@AronNovak I've attached the before and after video in the PR description. 

> Is it ported from a client project? Does it affect all of our projects with Focal Point module installed? 🤔

dipakmdhrm:
@AronNovak I'm realizing now you meant result of the crop in content. 
Here are screenshots:

Focal point|Before|After
-|-|-
focal_point|before|after

Issue #793, opened 2025-02-14
Replace google analytics with google_tag


mariano-dagostino:
Google Tag Manager (GTM) is preferred over directly using Google Analytics (GA) in Drupal because it allows easier management of tracking tags without modifying the site's code. You can add, update, or remove tags through its interface, reducing the need for developer involvement and minimizing errors. GTM also offers version control and debugging tools, making it more efficient for managing multiple tags and complex tracking setups.

https://www.drupal.org/project/google_tag

Differences:

https://www.analyticsmania.com/post/google-tag-manager-vs-google-analytics/

PR #794, opened 2025-02-14
Replace google_analytics with google_tag module

mariano-dagostino:
#793 

Screenshot_2025-03-03_11-39-45


amitaibu (approved):
Thanks, Screenshot(s) and MWG

PR #798, opened 2025-02-25
Replace bower/fontawesome with bower/font-awesome

mariano-dagostino:
Sometimes we see:

```
Problem 1
    - Root composer.json requires bower-asset/fontawesome == 6.4.2.0 (exact version match), found bower-asset/fontawesome[dev-update-readme-logo, dev-master, dev-issue-18807, v1.0.0, 2.0.0, v3.0.0, ..., v3.2.1, v4.0.0, ..., 4.x-dev, 5.0.6, ..., 5.x-dev, 6.0.0, ..., 6.x-dev] but these do not match your constraint and are therefore not installable. Make sure you either fix the constraint or avoid updating this package to keep the one present in the lock file (bower-asset/fontawesome[6.4.2]).
```

But it seems there are 2
    
- https://asset-packagist.org/package/bower-asset/fontawesome
- https://asset-packagist.org/package/bower-asset/font-awesome

Today they are the same, but we have seem some discrepancy in the past.

### DIFF
--- a/composer.patches.json
+++ b/composer.patches.json
@@ -16,9 +16,6 @@
     "drupal/entity_reference_revisions": {
       "d.o. #2799479": "https://www.drupal.org/files/issues/2022-06-01/entity_reference_revisions-relationship_host_id-2799479-176.patch"
     },
-    "drupal/google_analytics": {
-      "d.o #3373921": "https://www.drupal.org/files/issues/2023-08-07/google-analytics-issues-3373921-cannot-install-from-existing-config-11.patch"
-    },
     "nodespark/des-connector": {
       "php 8.1 compatibility": "patches/des-connector-php-compat.patch"
     },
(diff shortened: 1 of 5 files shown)


AronNovak (approved):
Oh the hyphen :)

robo-components/DeploymentTrait.php
```diff
@@ -86,6 +86,8 @@ trait DeploymentTrait {
     'package-lock.json',
     'composer.json',
     'composer.lock',
+    'web/libraries/font-awesome/js-packages',
```
amitaibu, line 89: Let's move above `web/theme/...`

composer.patches.json
```diff
@@ -16,9 +16,6 @@
     "drupal/entity_reference_revisions": {
       "d.o. #2799479": "https://www.drupal.org/files/issues/2022-06-01/entity_reference_revisions-relationship_host_id-2799479-176.patch"
     },
-    "drupal/google_analytics": {
```
amitaibu, line 19: This was committed?

amitaibu (commented):
Thanks

PR #799, opened 2025-02-25
Update to latest Drupal core with dependencies 

amitaibu:
```
ddev composer update "drupal/core-*" --with-all-dependencies
ddev drush cex -y
```

### DIFF
--- a/composer.lock
+++ b/composer.lock
@@ -1764,29 +1764,27 @@
         },
         {
             "name": "doctrine/deprecations",
-            "version": "1.1.3",
+            "version": "1.1.4",
             "source": {
                 "type": "git",
                 "url": "https://github.com/doctrine/deprecations.git",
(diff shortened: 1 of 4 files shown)


amitaibu:
@AronNovak before I dig in, maybe you'd remember: When executing https://drupal-starter.ddev.site:4443/search?key[$testing]=1 I get

image

When I try on `main`, it's the same behavior. So I'm wondering how this passed on `main`

https://github.com/Gizra/drupal-starter/blob/3560593128d317689f92ebbd45d1c5270e7e8d04/web/modules/custom/server_general/tests/src/ExistingSite/ServerGeneralSearchTest.php#L130

AronNovak:
@amitaibu I am sure at a certain point, the behavior of `core` has been changed. Previously there was a watchdog error message, now there's a WSOD. I adapted the test. The whole point is to make sure that this call does not fully render the server unusable.

PR #800, opened 2025-02-25
WIP: PHPStan rule to prevent `#theme` inside PEVB

amitaibu:
The goal is to get 

```php
<?php

namespace Drupal\server_general\Plugin\EntityViewBuilder;

use Drupal\Core\Entity\EntityViewBuilder;

class TestEntityViewBuilder extends EntityViewBuilder {
    public function build(array $build) {
        $build['#theme'] = 'test_theme'; // Should trigger the rule
        return $build;
    }
}
```

Produce something like 

```
Line   web/modules/custom/server_general/src/Plugin/EntityViewBuilder/TestEntityViewBuilder.php
----   -----------------------------------------------------------------------------------
10     Use of '#theme' is not allowed in EntityViewBuilder classes
```

### DIFF
--- a/phpstan.neon
+++ b/phpstan.neon
@@ -1,6 +1,7 @@
 includes:
   - vendor/mglaman/phpstan-drupal/extension.neon
   - vendor/phpstan/phpstan-deprecation-rules/rules.neon
+  - web/modules/custom/server_general/phpstan-extension.neon
 parameters:
   level: 6
   reportUnmatchedIgnoredErrors: false
--- a/web/modules/custom/server_general/phpstan-extension.neon
+++ b/web/modules/custom/server_general/phpstan-extension.neon
@@ -1,0 +1,5 @@
+services:
+    -
+        class: Drupal\server_general\PHPStan\NoThemeInEntityViewBuilderRule
+        tags:
+            - phpstan.rules.rule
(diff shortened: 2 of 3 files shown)


amitaibu:
Currently getting this error

image

balagan73:
@amitaibu classes in the PHPStan namespace cannot be found by the IDE (VScode in my case). PHPStan exists as a phar file, and chatgpt says

> "Since PHPStan exists as a .phar in your setup, it won’t automatically use Drupal’s autoloading. You need to explicitly provide the correct autoload file when running PHPStan."

I hope it helps.

amitaibu:
This is now working.

```php
// Will trigger error.
 $element['#theme'] = 'foo';

// Will trigger error.
$element = ['#theme' => 'foo'];

// No error, since it's a comment.
// $element = ['#theme' => 'foo'];
```

Example if having `#theme` inside a PEVB plugin

image

composer.json
```diff
@@ -125,6 +125,11 @@
             "RoboComponents\\": "./robo-components/"
         }
     },
+    "autoload-dev": {
```
bboro, line 128: Should we move `"RoboComponents\\": "./robo-components/"` above to `autoload-dev` too? IIRC we'll only need robo on ddev.

bboro (approved):

PR #801, opened 2025-02-26
Consolidate `ThemeTraits` and add PHPStan rule

amitaibu:
- Moved all `ThemeTraits` under a single directory
- Rename classes to have a `ThemeTrait` suffix
- PHPStan rule ensures there is no function with arguments of type `EntityInterface`

So for example if under `ThemeTrait\ButtonTrait`  we'd have

```php
  protected function buildLinkButton(ContentEntityInterface $entity, string $field_name = 'field_link'): array {
    return [];
  }
```

Then we get an error 

image


amitaibu:
image

amitaibu:
@mariano-dagostino ^^

web/modules/custom/server_general/src/ButtonTrait.php
```diff
@@ -1,55 +1,21 @@
...
-use Drupal\Core\StringTranslation\TranslatableMarkup;
-use Drupal\Core\Url;
 use Drupal\file\FileInterface;
 use Drupal\pluggable_entity_view_builder\BuildFieldTrait;
 
 /**
- * Helper methods for getting a themed button.
+ * Helper methods for getting a button from content.
  */
 trait ButtonTrait {
```
amitaibu, line 14: I'm not sure about the name. As now we have a `ThemeTrait\ButtonTrait` and a "regular" `ButtonTrait`.

Ideas are welcome.

1. Should we maybe remove the `Trait` suffix from the traits under `ThemeTrait`?
2. OR maybe call this one `ButtonHelperTrait`?

amitaibu: I've renamed the `ThemeTrait` classes

Issue #807, opened 2025-03-03
Use and Document Pantheon's new way to set secret envs [2h]


amitaibu:
https://github.com/Gizra/drupal-starter/blob/d34921d6d3ce4fd9ac9ac104643f4e860273dcb0/web/sites/default/settings.pantheon.php#L130-L135

This way, but this is [deprecated](https://docs.pantheon.io/guides/secrets#older-plugin-now-deprecated).

This is the [new way](https://docs.pantheon.io/guides/secrets/php#use-the-pantheon_get_secret-php-function):
```php
if ( function_exists('pantheon_get_secret') ) {
  $secret_value = pantheon_get_secret('secret_name');
}
```

We should refactor 2FA and document it for AI too.
It could be a separate task.

_Originally posted by @AronNovak in https://github.com/Gizra/drupal-starter/pull/802#discussion_r1977013556_

AronNovak:
@amitaibu Should I create a documentation PR for covering the AI module too?

amitaibu:
That would be great, thanks!

PR #810, opened 2025-03-04
Better secrets handling for Pantheon

AronNovak:
#807 

## Status

```
gizra  drupal-starter  terminus remote:drush gizra-drupal-starter.qa php:eval 'print pantheon_get_secret("tfa_key");'
 [warning] This environment is in read-only Git mode. If you want to make changes to the codebase of this site (e.g. updating modules or plugins), you will need to toggle into read/write SFTP mode first.
12345678910XXXXXXXXXXXXXXXXXXXXXXXXX/Cs= [notice]
Command: gizra-drupal-starter.qa -- drush php:eval print pantheon_get_secret("tfa_key"); [Exit: 0] (Attempt 1/1)
```

### DIFF
--- a/.ddev/commands/web/terminus
+++ b/.ddev/commands/web/terminus
@@ -4,18 +4,26 @@
 ## Usage: terminus [flags] [args]
 ## Example: "ddev terminus remote:drush drupal-starter.live uli"
 
-# Function to check if the terminus-secrets-manager-plugin is installed.
+# Function to check if a specified plugin is installed.
 is_plugin_installed() {
-  terminus self:plugin:list | grep -q "terminus-secrets-manager-plugin"
+  local plugin_name="$1"
+  terminus self:plugin:list | grep -q "$plugin_name"
 }
 
-# Check if the first argument is in the "secret" namespace.
-if [[ $1 == secret:* ]]; then
-  # Check if the plugin is not installed
-  if ! is_plugin_installed; then
-    echo "Installing terminus-secrets-manager-plugin..."
-    terminus self:plugin:install terminus-secrets-manager-plugin
+# Function to install a specified plugin if it's not installed.
+install_plugin_if_missing() {
+  local plugin_name="$1"
+  if ! is_plugin_installed "$plugin_name"; then
+    echo "Installing $plugin_name..."
+    terminus self:plugin:install "$plugin_name"
   fi
+}
+
+# Secret Manager is rarely needed, this way we do not install it
+# again and again at the CI environment for instance.
+# Only when it's actually needed, we install it on-the-fly.
+if [[ $1 == secret:* ]]; then
+  install_plugin_if_missing "terminus-secrets-manager-plugin"
 fi
 
 # Run the terminus command with all provided arguments
--- a/.ddev/commands/web/terminus
+++ b/.ddev/commands/web/terminus
@@ -1,0 +1,7 @@
+#!/bin/bash
+
+## Description: Run terminus inside the web container
+## Usage: terminus [flags] [args]
+## Example: "ddev terminus remote:drush drupal-starter.live uli"
+
+terminus $@
--- a/.ddev/commands/web/terminus
+++ b/.ddev/commands/web/terminus
@@ -4,4 +4,19 @@
 ## Usage: terminus [flags] [args]
 ## Example: "ddev terminus remote:drush drupal-starter.live uli"
 
-terminus $@
+# Function to check if the terminus-secrets-manager-plugin is installed.
+is_plugin_installed() {
+  terminus self:plugin:list | grep -q "terminus-secrets-manager-plugin"
+}
+
+# Check if the first argument is in the "secret" namespace.
+if [[ $1 == secret:* ]]; then
+  # Check if the plugin is not installed
+  if ! is_plugin_installed; then
+    echo "Installing terminus-secrets-manager-plugin..."
+    terminus self:plugin:install terminus-secrets-manager-plugin
+  fi
+fi
+
+# Run the terminus command with all provided arguments
+terminus "$@"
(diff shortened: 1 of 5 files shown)


.ddev/config.local.yaml.example
```diff
@@ -9,6 +9,9 @@ hooks:
     # Private files directory.
     - exec: mkdir -p /var/www/private
 
+    # Terminus secrets plugin.
```
amitaibu, line 12:
```suggestion
    # Install the Terminus secrets plugin.
```

.ddev/config.yaml
```diff
@@ -266,8 +266,12 @@ corepack_enable: false
 # and you can't erase existing hooks or all environment variables.
 # However, with "override_config: true" in a particular config.*.yaml file,
 # 'use_dns_when_possible: false' can override the existing values, and
-# hooks:
-#   post-start: []
+#
+hooks:
+  post-start:
+    # Terminus Secrets plugin.
+    - exec: terminus self:plugin:install terminus-secrets-manager-plugin
```
amitaibu, line 273: Do we need this? We have it also on .ddev/config.local.yaml.example

README.md
```diff
@@ -529,13 +529,23 @@ To import the config translations:
...
 login, they will already be blocked. A site admin may reset their validation tries
 under the `/admin/people` page.
 The TFA method that is enabled is one that uses Google authenticator (or similar).
 
+You should set the TFA secret using:
+```bash
+ddev terminus secret:site:set gizra-drupal-starter tfa_key $(openssl rand -base64 32) --type=runtime --scope=web,user
+```
+
+If you need environment-based override, you can do the following:
```
amitaibu, line 544: What's the diff between this command and the one above?

amitaibu (commented):
Thanks

README.md
AronNovak: The `.qa` suffix - specifying where the value should be applied.

.ddev/config.yaml
AronNovak: The problem is that `.ddev/config.local.yaml.example` is might or might not be copied to the proper place. Let's say you'd like to quickly start the project just to do some non-site specific thing. It happens for me often. So it is to make sure the plugin is **always** installed. We cannot put it in the `Dockerfile`, as the installation is per-user, the `Dockerfile` installs it for `root`.

README.md
amitaibu:
```suggestion
If you need to override the secret on a specific Pantheon environment:
```

.ddev/config.yaml
amitaibu: Or actually: let's keep only on `.ddev/config.local.yaml.example`. I treat not copying the local yaml as an edge case

.ddev/config.yaml
```diff
@@ -266,8 +266,12 @@ corepack_enable: false
 # and you can't erase existing hooks or all environment variables.
 # However, with "override_config: true" in a particular config.*.yaml file,
 # 'use_dns_when_possible: false' can override the existing values, and
-# hooks:
-#   post-start: []
+#
+hooks:
+  post-start:
+    # Terminus Secrets plugin.
```
AronNovak, line 272:
```suggestion
    # Install the Terminus secrets plugin.
```

.ddev/config.yaml
AronNovak: The problem is that it does not survive a `ddev stop` and `ddev start` cycle. And that's not an edge case, we do stop DDEV projects to gain back resources and then start it to continue the development. And there `.ddev/config.local.yaml.example` is harmful as it wipes the database.

`$HOME` is ephemeral, it is thrown away immediately when the container is stopped. Only the image is persisent.

```
aaron@drupal-starter-web:/var/www/html$ find ~/.terminus/ -type d | head
/home/aaron/.terminus/
/home/aaron/.terminus/terminus-dependencies-a6b56bb299
/home/aaron/.terminus/terminus-dependencies-a6b56bb299/vendor
/home/aaron/.terminus/terminus-dependencies-a6b56bb299/vendor/guzzlehttp
/home/aaron/.terminus/terminus-dependencies-a6b56bb299/vendor/guzzlehttp/psr7
/home/aaron/.terminus/terminus-dependencies-a6b56bb299/vendor/guzzlehttp/psr7/src
/home/aaron/.terminus/terminus-dependencies-a6b56bb299/vendor/guzzlehttp/psr7/src/Exception
/home/aaron/.terminus/terminus-dependencies-a6b56bb299/vendor/guzzlehttp/promises
/home/aaron/.terminus/terminus-dependencies-a6b56bb299/vendor/guzzlehttp/promises/src
/home/aaron/.terminus/terminus-dependencies-a6b56bb299/vendor/guzzlehttp/guzzle
```

And that's where this plugin is installed.

I will check if we can still somehow use the DDEV image to contain the plugin, it would be the best.

amitaibu: I see, you want to ensure that the post-start always has it. That's fine, then we can leave as is.

But I guess the termins secret plugin is needed quite rarely. So if one would  `ddev terminus ... ` and won't get it working, they could figure it was not installed.

AronNovak: See the [last commit](https://github.com/Gizra/drupal-starter/pull/810/commits/79ddf2c062307cb013fb2febac886013831939b5), I decided to install it on-the-fly, so we won't download it thousands of times as week via Travis and in other scenarios, when it's not needed. What do you think?

.ddev/commands/web/terminus
```diff
@@ -0,0 +1,22 @@
...
+## Description: Run terminus inside the web container
+## Usage: terminus [flags] [args]
+## Example: "ddev terminus remote:drush drupal-starter.live uli"
+
+# Function to check if the terminus-secrets-manager-plugin is installed.
+is_plugin_installed() {
+  terminus self:plugin:list | grep -q "terminus-secrets-manager-plugin"
+}
+
+# Check if the first argument is in the "secret" namespace.
```
amitaibu, line 12: :ok_hand: 
Let's add a comment explaining the incentive behind this.

amitaibu (approved):
LGTM

PR #816, opened 2025-03-14
Dynamic cache setting for PHPunit vs local development

bboro:


bboro:
0.2h

web/sites/default/settings.ddev.php
```diff
@@ -61,6 +61,24 @@
 // better performance.
 $settings['class_loader_auto_detect'] = FALSE;
 
+// Cache settings for local development. Below enables caching during testing,
+// and disables caching at other times on ddev which is handy during develoment.
+// If you wish to override below settings, you can do so in
+// /web/sites/default/settings.local.php (not /web/sites/settings.local.php!).
+$config['system.performance']['css']['preprocess'] = defined('DTT_BASE_URL');
```
AronNovak, line 68: Why not relying on https://docs.travis-ci.com/user/environment-variables, `CI=true` for instance?
So inside Travis, it's always enabled, regardless of the method.
We need to make sure we try to pass `CI=true` like `TRAVIS_COMMIT_MESSAGE`.

PR #817, opened 2025-03-18
Pantheon - easy remote access to services

AronNovak:


AronNovak:
> Should we include in this trait the DDOS protection to unify all pantheon related stuff?

According to my taste, these are two distinct topics. @amitaibu What do you prefer here?

amitaibu (approved):
I think you are both right :)

Since it's already like this, let's go ahead and get it in.

PR #819, opened 2025-03-18
PHPstan rule to detect incorrect use of addCacheableDependency

mariano-dagostino:


AronNovak (approved):

AronNovak:
I generally wonder if we should add some test coverage for our custom PphStan rules to keep them working on the long run.

amitaibu:
💯 

I suggest, to get this in, and create a PR against https://github.com/mglaman/phpstan-drupal adding this. Let's see if we can move this upstream.

amitaibu:
@mariano-dagostino I briefly spoke with @mglaman regarding this PR. It can be ported to https://github.com/mglaman/phpstan-drupal because it's a great idea!:smile:

amitaibu:
@mariano-dagostino reminder ^^

Issue #820, opened 2025-03-20
Update Core + Contrib [1h]


amitaibu:
<img width="1311" alt="Image" src="https://github.com/user-attachments/assets/759f98d6-aa65-4afc-b6ed-3d5886e9e5f7" />

PR #828, opened 2025-04-15
Update core and contrib

balagan73:
#820 

- Updated drupal core to 10.4.6.
- Updated contrib modules.
- Ran drush updb, drush cex, there were no config changes.

TB: 0.3/1h


amitaibu (approved):
:+1:

PR #836, opened 2025-05-01
Execute deploy hooks

amitaibu:
Spotted on a client's project

### DIFF
--- a/.ddev/config.local.yaml.example
+++ b/.ddev/config.local.yaml.example
@@ -44,6 +44,9 @@ hooks:
     # @see https://www.drupal.org/node/947312
     - exec: drush user-block --uid=1
 
+    # Run deploy hooks.
+    - exec: drush deploy:hook --no-interaction
+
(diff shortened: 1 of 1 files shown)


.ddev/config.local.yaml.example
```diff
@@ -56,8 +59,8 @@ hooks:
     - exec-host: ddev login
 
   post-import-db:
-    # Run DB updates.
-    - exec: drush updb
+    # Run deploy hooks.
+    - exec: drush deploy:hook --no-interaction
```
AronNovak, line 63: We should preserve `drush updb`, shouldn't we?
Also this should be the last step, after config import, see the [Robo logic](https://github.com/Gizra/drupal-starter/blob/main/robo-components/DeploymentTrait.php#L418).

AronNovak:
Then should we adapt DeploymentTrait.deployPantheonSync too to do the same logic?

PR #845, opened 2025-05-15
Fix default value of wrapper gap

amitaibu:
| Before | After |
| -- | -- |
| image | image |


amitaibu:
Trivial PR, self-merging

PR #866, opened 2025-06-20
Entity query - latest revision performance fix

AronNovak:
https://www.drupal.org/node/2950869

We were bitten by this at a larger client project - using Scheduler module. But the problem is generic, entity queries are affected generally.

### DIFF
--- a/composer.patches.json
+++ b/composer.patches.json
@@ -4,7 +4,8 @@
       "d.o. #3315245": "patches/2938.patch",
       "d.o. #2351685": "https://www.drupal.org/files/issues/2023-02-03/2351685-22.patch",
       "d.o. #3400204": "https://www.drupal.org/files/issues/2024-06-27/3400204-claro-layout-overflow-fix-2.patch",
-      "d.o. #3456176: Fix Messages rendering with bigpipe": "https://www.drupal.org/files/issues/2024-06-28/3456176-45-missing-status-messages-theme-suggestions.patch"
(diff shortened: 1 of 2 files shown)


composer.patches.json
```diff
@@ -4,7 +4,8 @@
       "d.o. #3315245": "patches/2938.patch",
       "d.o. #2351685": "https://www.drupal.org/files/issues/2023-02-03/2351685-22.patch",
       "d.o. #3400204": "https://www.drupal.org/files/issues/2024-06-27/3400204-claro-layout-overflow-fix-2.patch",
-      "d.o. #3456176: Fix Messages rendering with bigpipe": "https://www.drupal.org/files/issues/2024-06-28/3456176-45-missing-status-messages-theme-suggestions.patch"
+      "d.o. #3456176: Fix Messages rendering with bigpipe": "https://www.drupal.org/files/issues/2024-06-28/3456176-45-missing-status-messages-theme-suggestions.patch",
+      "d.o. #2950869": "patches/core-query-revision-performance.patch"
```
amitaibu, line 8: Thanks, we don't need the local copy, lets point to the d.o url

amitaibu (approved):

composer.patches.json
AronNovak: image

These are MRs, so the URL is a moving target, that's why I copied it.

amitaibu: Ok

PR #867, opened 2025-06-23
Add automatic updater robo component.

balagan73:
Most updates consists of simple repetitive tasks. Trying to execute these automatically.


balagan73:
https://github.com/user-attachments/assets/6af5e9e4-4a69-4cb5-8e31-4a07a6620d2a

robo-components/AutoUpdateTrait.php
```diff
@@ -0,0 +1,53 @@
...
+   * Update drupal core and modules.
+   *
+   * @throws \Exception
+   */
+  public function updateModules() {
+    if (\Drupal::moduleHandler()->moduleExists('update')) {
+      $this->say("Update module is installed, checking status of projects.");
+      if ($available = update_get_available(TRUE)) {
+        \Drupal::moduleHandler()->loadInclude('update', 'compare.inc');
+        // @phpstan-ignore-next-line
```
AronNovak, line 22: Can we get rid of phpstan ignore?

AronNovak: Well, it should know about it, see https://github.com/mglaman/phpstan-drupal/issues/124 and https://github.com/mglaman/phpstan-drupal/issues/166, if nothing works, we should file a new bugreport. It is kind of out of scope here, so let's check if it works, if not, open an issue and we can go on.

robo-components/AutoUpdateTrait.php
```diff
@@ -0,0 +1,55 @@
...
+      $exit_code = $this->taskExec("composer update 'drupal/" . $package . ":^" . $project['recommended'] . "' -W")
+        ->printOutput(TRUE)
+        ->run()
+        ->getExitCode();
+      if ($exit_code !== 0) {
+        $this->yell('There was an error updating ' . $package . ' to version ' . $project['recommended']);
+      }
+      // Update successful, add composer.lock to staging area,
+      // then commit it.
+      $this->taskExec("git add composer.lock")->printOutput(TRUE)->run();
```
AronNovak, line 49: We do this even if no package could be updated successfully above.

PR #873, opened 2025-07-23
Better pantheon remote cmds

AronNovak:
From a client project, it turned out that there's room it improve. If you can't pipe something into a mysql console, it does not worth too much :)

### DIFF
--- a/robo-components/PantheonRemoteTrait.php
+++ b/robo-components/PantheonRemoteTrait.php
@@ -28,8 +28,7 @@ trait PantheonRemoteTrait {
 
     $sftp_command = trim($result->getMessage());
 
-    // Execute the SFTP command.
-    $this->_exec($sftp_command);
+    $this->executeCommand($sftp_command);
   }
 
   /**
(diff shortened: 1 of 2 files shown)


robo-components/PantheonRemoteTrait.php
```diff
@@ -126,4 +125,22 @@ public function pantheonConnectInfo(string $env = 'dev', string $type = 'all'):
...
+   * Execute a command using the appropriate method based on stdin status.
+   *
+   * Uses passthru() when input is piped (for stdin support) or _exec() for
+   * interactive commands (for better Robo integration and output formatting).
+   */
+  private function executeCommand(string $command): void {
+    // Check if stdin is being piped.
+    if (!posix_isatty(STDIN)) {
+      // Input is piped, use passthru for stdin support.
+      passthru($command);
```
amitaibu, line 138: When is this going to be called?

AronNovak: `echo "select * from node" | ddev robo ...`

Without this PR, it does not work.

amitaibu: Let's add it as a comment, and MWG

PR #885, opened 2025-08-14
Re-add Selenium files

amitaibu:
Re-ran `ddev add-on get ddev/ddev-selenium-standalone-chrome`


.ddev/config.selenium-standalone-chrome.yaml
```diff
@@ -1,3 +1,7 @@
+#ddev-generated
```
amitaibu, line 1: This is needed, so new version can override the file

PR #887, opened 2025-08-19
WIP: Dump cache of Content view when main_settings is updated

bboro:


bboro:
Attempted to use https://www.drupal.org/project/views_custom_cache_tag but it made no difference here with the correct configuration. Only setting the caching to "None" seems to work.

TB: 0.7 / 1h

PR #907, opened 2025-09-18
Fix semver version in updater

balagan73:
We need to use semver, instead of the legacy drupal versioning drupal provides on the status page.


AronNovak (approved):

balagan73:
image

Git log after running the script

image

AronNovak:
@balagan73 Looks great!

PR #908, opened 2025-09-22
End of file newline pre-commit hook

AronNovak:
## Status
image

### DIFF
--- a/.ddev/git/hooks/pre-commit
+++ b/.ddev/git/hooks/pre-commit
@@ -43,4 +43,3 @@ fi
 
 echo "All staged text files have proper newlines at the end."
 exit 0
-


AronNovak:
@mariano-dagostino Pre-commit hooks cannot be shared trivially with the whole team, otherwise it is a great idea, I went ahead, but this is a limitation.
Perhaps `ddev start` could move to the right place...

AronNovak:
This almost only a problem for AI generated code.
https://www.drupal.org/docs/develop/standards/php/php-coding-standards
> All text files should end in a single newline (\n). This avoids the verbose "\ No newline at end of file" patch warning and makes patches easier to read since it's clearer what is being changed when lines are added to the end of a file.

That's why we care. Editors and IDEs take care due to:
```
root = true

# All files.
[*]
end_of_line = LF
indent_style = space
indent_size = 2
charset = utf-8
trim_trailing_whitespace = true
insert_final_newline = true

[composer.{json,lock}]
indent_size = 4
```

So it only happens for generated files. Likely better to adapt the initial prompt to somehow enforce stuff from `editorconfig`.

PR #914, opened 2025-10-01
Revert "Port phpcs parallel processing support (#900)"

AronNovak:
This reverts commit 969033b25266bb76560450703b52f43df11fa44e.

On client projects, we had mixed results, it caused slowdowns.


AronNovak:
Self-merging as it was agreed at another discussion and it's just a revert commit.

PR #918, opened 2025-10-21
Improve auto update script

balagan73:
- Check if module is compatible with current core version.
- Run update in a separate subprocess to avoid errors due to composer update removing /vendor contents.

TB: 0.3

### DIFF
--- a/robo-components/AutoUpdateTrait.php
+++ b/robo-components/AutoUpdateTrait.php
@@ -70,7 +70,8 @@ trait AutoUpdateTrait {
       // Update core in a separate subprocess, as otherwise composer
       // might get updated too, resulting in deleted files inside
       // the /vendor directory causing an error.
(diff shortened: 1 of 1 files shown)


robo-components/AutoUpdateTrait.php
```diff
@@ -39,6 +39,12 @@ public function updateModules() {
         // No need to update.
         continue;
       }
+      // If recommended module is not compatible with current drupal
+      // core, we skip the update.
+      if (isset($project['releases'][$project['recommended']]['core_compatible']) && $project['releases'][$project['recommended']]['core_compatible'] === FALSE) {
+        $this->say($project['recommended'] . ' is not compatible with current drupal core, skipping it.');
```
AronNovak:
```suggestion
        $this->say($project['name'] . ' (version ' . $project['recommended'] . ') is not compatible with current drupal core, skipping it.');
```

Issue #920, opened 2025-10-24
No h1 in WYSIWYG [1h]


isapisa:
### Issue

To prevent having duplicate `h1` tags on a page, we shouldn't allow adding heading 1 in the WYSIWYG:

<img width="1246" height="668" alt="Image" src="https://github.com/user-attachments/assets/1fd044a5-3950-4354-b7c6-ffdad3beb311" />

### To Do

Remove `Heading 1` by default from WYSIWYG editor

PR #925, opened 2025-10-24
No h1 in WYSIWYG

balagan73:
#920 

- Remove H1 tag from full html text format.

TB: 0.2/1h


balagan73:
Basic HTML|Full HTML
-|-
image|image

Issue #924, opened 2025-10-24
Configure meta name description and title for content types and allow override [3h]


isapisa:
### Issue
`<meta name="description">` and `<title>` is used by search engines, so we need to:
- Make sure it is configured for each content type
- Allow customisation for each content type

By default:
- `<title>` = `Node title|[site:name]`
- `<meta name="description">` = `summary` or `first 300 chars of description/text paragraph`

Customisation:

<img width="896" height="1272" alt="Image" src="https://github.com/user-attachments/assets/0a321b85-f4f9-4c8d-aaf9-d13413a99852" />

### To Do
- Configure defaults for `<title>` and `<meta name="description">` for all content types
- Implement customisation for all content types

PR #926, opened 2025-10-26
Configure meta name description and title for content types and allow override

balagan73:
#924 

- Allow costumization of metatags for all CTs.

TB: 1.1/3h


balagan73:
Landing page|News
-|-
image|image

@isapisa I have enabled costumization for both CTs we have here. I did not set any overrides though. They both inherit the required settings from Content:
title:	[node:title] | [site:name]
description:	[node:summary]

The node:summary is irrelevant for the Landing page, as it has no body field, but field_paragraphs, and we cannot know in advance what paragraphs will be used. User can customize it though if needed.

PR #927, opened 2025-10-26
Rename server_general_node_access to server_general_entity_access

amitaibu:
I don't think `hook_node_access` still exists on D10


amitaibu:
hmm, I see `hook_node_access` is invoked, but the $op passed is `view all revisions` instead of `view`

PR #937, opened 2025-11-05
Print messages in the header

amitaibu:
image


bboro (approved):

amitaibu:
Actually before merging, I'll write a test, to ensure it doesn't happen again

PR #944, opened 2025-11-21
Fail phpunit parallel testing if a test fails

amitaibu:
See example where tests would pass, if there's no `set -e` present

https://github.com/amitaibu/Microsites-Drupal-Starter/pull/8/commits/89d3e95ae4d6235d94929b8a2f12fe03cea7eb11


amitaibu:
@AronNovak, this is now working correctly, as we have failing tests, and it no longer reports as passing.

image

PR #949, opened 2025-11-25
Update core and modules

balagan73:
#941 

- Update drupal core to 10.5.6.
- Update contrib modules.

TB: 0.4/2h


AronNovak (approved):
Looks good

PR #958, opened 2025-12-16
Manage access to Styleguide with permissions

bboro:


bboro:
Ported from another project.

Issue #922, opened 2025-10-24
Avoid duplicate canonical tags and soft redirects [2h]


isapisa:
### Issue
To improve SEO, we need to make sure that:
- `Zero second Soft redirect` is not configured as it might be confusing the crawler.
- We don't have duplicate `canonical tags`.

### To Do
- Replace `No 0-sec refresh` with a link
- Avoid duplicate canonical tag

balagan73:
@isapisa This is more complicated than I thought. The canonical tag issue is actually a pevb bug/feature. There we take direct control of the rendered output, and not let other modules to alter it, that's why metatag cannot remove duplicate canonicals. I am out of TB and need some more time to create a fix/patch for pevb.

balagan73:
This should fix the problem: https://www.drupal.org/project/pluggable_entity_view_builder/issues/3554759
Currently solution requires metatag patch too: https://www.drupal.org/files/issues/2025-10-29/metatag-nullable-display.patch

PR #959, opened 2025-12-17
Update core and modules

balagan73:
#922 

- Update drupal core
- Update contrib modules
- Drush updb, cex.

TB: 0.5h


balagan73:
drush updb
image

PR #980, opened 2026-01-29
Support phpactor

mariano-dagostino:
https://phpactor.readthedocs.io/en/master/integrations/drupal8.html

https://github.com/user-attachments/assets/a96ccf0e-ebac-460d-896e-b1c382df22d5


AronNovak (commented):
I see that there's phpstan integration. Did you use it? Can it auto-correct the findings using `phpactor`?

composer.json
```diff
@@ -186,6 +188,17 @@
                 "web/modules/contrib/webform/composer.libraries.json"
             ]
         },
+        "extra": {
+            "autoload-drupal": {
+                "modules": [
+                    "app/modules/contrib/",
```
AronNovak: https://phpactor.readthedocs.io/en/master/integrations/drupal8.html#option-1-bootstrap-drupal-on-the-fly-to-generate-the-autoloader

I see that there are two ways to use it. This is option 2). It seems to be more invasive. But might be fine, just asking why option 1) isn't chosen here.

AronNovak (approved):

AronNovak (approved):

amitaibu:
Let's put this on hold for now. I don't see much value in it, in the AI era

PR #992, opened 2026-02-11
Migrate DDEV theme-watch command from PostCSS to Tailwind CLI

wombatbuddy:
After migrating to Tailwind v4, PostCSS is no longer used for CSS compilation. This PR updates the `theme-watch` command to use Tailwind CLI instead.


amitaibu (approved):
Thanks 👍

PR #999, opened 2026-03-09
Git worktree support

AronNovak:
https://docs.ddev.com/en/stable/users/configuration/config/#omit_project_name_by_default

https://www.fldrupal.camp/session/use-git-worktree-ddev-run-multiple-versions-same-site


amitaibu (approved):
Oh, seems we never merged it

PR #1011, opened 2026-03-20
Fix XSS risk in QuoteThemeTrait

AronNovak:
## Summary
- Use `#plain_text` instead of `#markup` for image credit field value in QuoteThemeTrait
- Prevents stored XSS via unsanitized `field_media_credit` content

## Test plan
- [ ] Verify quote paragraphs still render image credit correctly

### DIFF
--- a/web/modules/custom/server_general/src/ThemeTrait/QuoteThemeTrait.php
+++ b/web/modules/custom/server_general/src/ThemeTrait/QuoteThemeTrait.php
@@ -47,7 +47,7 @@ trait QuoteThemeTrait {
     // The photo credit on top of the image.
     $credit = [];
     if (!empty($image_credit)) {
-      $credit[] = ['#markup' => '© ' . $image_credit];
+      $credit[] = ['#plain_text' => '© ' . $image_credit];
     }
 
     return [


amitaibu:
Thanks, screenshot please

web/modules/custom/server_general/src/ThemeTrait/QuoteThemeTrait.php
```diff
@@ -47,7 +47,7 @@ protected function buildElementQuote(array $image, array $quote, ?string $subtit
     // The photo credit on top of the image.
     $credit = [];
     if (!empty($image_credit)) {
-      $credit[] = ['#markup' => '© ' . $image_credit];
+      $credit[] = ['#plain_text' => '© ' . $image_credit];
```
mariano-dagostino, line 50: I think maybe is better to specify `#allowed_tags`, strong and em tags are sometimes useful in credit sections. 

Also, according to [the docs](https://api.drupal.org/api/drupal/core%21lib%21Drupal%21Core%21Render%21theme.api.php/group/theme_render/11.x)

> #markup: Specifies that the array provides HTML markup directly. Unless the markup is very simple, such as an explanation in a paragraph tag, it is normally preferable to use #theme or #type instead, so that the theme can customize the markup. Note that the value is passed through [\Drupal\Component\Utility\Xss::filterAdmin](https://api.drupal.org/api/drupal/core%21lib%21Drupal%21Component%21Utility%21Xss.php/function/Xss%3A%3AfilterAdmin/11.x)(), which strips known XSS vectors while allowing a permissive list of HTML tags that are not XSS vectors. (For example, <script> and <style> are not allowed.) See [\Drupal\Component\Utility\Xss](https://api.drupal.org/api/drupal/core%21lib%21Drupal%21Component%21Utility%21Xss.php/class/Xss/11.x)::$adminTags for the list of allowed tags. If your markup needs any of the tags not in this list, then you can implement a theme hook and/or an asset library. Alternatively, you can use the key #allowed_tags to alter which tags are filtered.

Which type of XSS issue are we seeing here?

mariano-dagostino (changes requested):

web/modules/custom/server_general/src/ThemeTrait/QuoteThemeTrait.php
AronNovak: Right, keep it as-is. `#markup` is already safe.

PR #1016, opened 2026-03-31
Drupal canvas

andreytroeglazov:


andreytroeglazov:
[vokoscreenNG-2026-03-31_17-22-39.webm](https://github.com/user-attachments/assets/ccbb4441-7f83-48f7-ba78-075671f90d63)

Canvas module provides it's own type of pages, when creating a page it uses default front theme(we can configure layout and sections there), it allows adding blocks to the page, create and add custom components(I have added 2 examples) with fields. I have replicated CTA paragraph as sdc and reused wrappers, but it's reused in twig template, so it's the same as we have in php traits but in twig, haven't tried nested components or reference fields.
It looks ok and can be used, but it's slow for me and I have some js errors, also lost some time to find add component button to page, because it was hidden under leave button.
 
Selection_095

Probably we can reuse trait methods inside preprocesses instead of rendering everything in twig.

TB: 3.4h

@amitaibu ^^

PR #1018, opened 2026-04-02
AI content with paragraphs generation

andreytroeglazov:
Generated summary with claude of what I have discovered:

## Summary of Evaluated Approaches
                              
  ### Module-Based Approaches            
                             
  #### 1. `drupal/ai` + AI Automators (submodule)
                             
  **What it does:** Configures chained, multi-step field population triggered on entity save. No code required.
                                
  **Why not:** Designed for populating *existing fields* on save, not for deciding *which paragraph types to create and how many*. Our use case requires structural decision-making — the AI must choose which paragraphs to generate and in what order. AI Automators can't orchestrate multi-entity creation with parent-child relationships.                                                                            
                                                                                    
  #### 2. `drupal/ai_agents` + `drupal/paragraphs_ai`                                                                                                                                                                                                                                                                                                                                                                      
                                                                                                                                                                                                
  **What it does:** Plugin-based agent framework for text-to-action operations. Built-in agents create/edit content types, fields, taxonomies, and paragraph types. Provides a tool-use loop where the AI reasons and takes actions iteratively.                                                                                                                                                                           
                                                                                                                                                                                                                                                                                                                                                                                                                           
  **Why not for full generation:** These are *site-building* agents — they manage configuration (create paragraph types, add fields), not content. `paragraphs_ai` explicitly states: "This agent can not create or edit paragraph content — only the paragraph types themselves." The agent loop also adds unnecessary latency (3-5 API round-trips) for what is essentially a single-shot generation task.               
                                                                                   
  **What we will use from it:** `paragraphs_ai`'s `GetParagraphFieldInformation` plugin for *dynamic schema discovery* — so our generator automatically knows about paragraph types and their fields without hardcoding. This requires applying the patch from [#3580185](https://www.drupal.org/project/paragraphs_ai/issues/3580185) which adds the `GetParagraphFieldInformation`, `GetParagraphFieldConfigForm`, and   
  `CreateParagraphBundle` function-call plugins.                                   
                                                                                                                                                                                                                                                                                                                                                                                                                           
  #### 3. `drupal/ai_interpolator`                                                 
                                                                                                                                                                                                                                                                                                                                                                                                                           
  **What it does:** Predecessor to AI Automators — data transformation pipelines.  
                                
  **Why not:** AI Automators (already in `drupal/ai`) is its evolution. No advantage over what's already installed.
                                                                                                                                                                                                                                                                                                                                                                                                                           
  #### 4. `drupal/openai` (standalone module)
                                                                                                                                                                                                                                                                                                                                                                                                                           
  **What it does:** Original single-provider OpenAI integration with Drush content generation.
                                                                                                                                                                                                                                                                                                                                                                                                                           
  **Why not:** Community is shifting to `drupal/ai` (provider-agnostic). Would lock us to OpenAI only and duplicates functionality we already have.                                                                                                                                                                                                                                                                        
                           
  #### 5. `drupal/eca` + `drupal/ai_integration_eca`                                                                                                                                                                                                                                                                                                                                                                       
                                                                                   
  **What it does:** Event-Condition-Action workflow builder with AI actions.
                    
  **Why not:** Overkill for this task. Better suited for complex event-driven workflows, not single-shot content generation.                                                                                                                                                                                                                                                                                               
                             
  #### 6. Drupal Canvas AI                                                                                                                                                                                                                                                                                                                                                                                                 
                                                                                   
  **What it does:** Visual page builder with AI assistant for generating complete pages.                                                                                                                                                                                                                                                                                                                                   
                    
  **Why not:** Generates Canvas pages, not traditional nodes with paragraph entities. Different content architecture than our project uses.                                                                                                                                                                                                                                                                                
                                                                                   
  ### Chosen Approach: Custom Form + Service using `drupal/ai`
                          
  **What it does:** A custom module (`server_ai_content`) with an admin form at `/admin/content/ai-generate` where the user types a prompt. A `ParagraphSchemaDiscovery` service uses `paragraphs_ai`'s `GetParagraphFieldInformation` (from patch [#3580185](https://www.drupal.org/project/paragraphs_ai/issues/3580185)) to dynamically discover available paragraph types and their fields. A `ContentGenerator`       
  service sends one structured GPT-4o call (via `drupal/ai` provider abstraction) with the discovered schema + user prompt, parses the JSON response, creates paragraph entities (with DALL-E 3 images where needed), attaches them to a landing page node, and saves as draft. User is redirected to the unpublished node preview.
                                                                                                                                                                                                                                                                                                                                                                                                                           
  **Example prompt:**                                                              
                                
  > Create me a landing page for the 2026 New Year celebration. Include a hero section, event details, FAQ section, and a call to action for registration.
                                                                                                                                                                                                                                                                                                                                                                                                                           
  **Why this approach:**     
                                                                                                                                                                                                                                                                                                                                                                                                                           
  - **Simplest path**        
  - **Uses existing infrastructure** — `drupal/ai` for OpenAI calls (already configured), `paragraphs_ai` for schema discovery (with patch [#3580185](https://www.drupal.org/project/paragraphs_ai/issues/3580185))                                                                                                                                                                                                        
  - **Single API call** — GPT sees full page context, content flows naturally between sections, lower latency than agent loops
  - **Dynamic schema** — if paragraph types change, the generator picks it up automatically                                                                                                                                                                                                                                                                                                                                
  - **Full control** — we define exactly how entities are created and attached     
  - **POC-friendly** — synchronous flow with spinner, can be extended later (batch processing, more content types, provider swapping)


andreytroeglazov:
image
Result:
screencapture-drupal-starter-ddev-site-4443-world-rock-music-2026-04-02-16_19_16

Will push code soon, still testing.

TB: 2.4h investigation + 3.4h implementation

amitaibu:
Wow! :muscle:

amitaibu:
Later we can move this form to be on the top of `/node/add` -- so a content editor can decide to create manually, or just go with the form

andreytroeglazov:
screencapture-drupal-starter-ddev-site-4443-marvel-comics-epic-crossover-event-2026-04-02-17_53_47

Fixed some problems, tuned promt a bit.

@amitaibu right now it uses only paragraphs without references to other content on the site

andreytroeglazov:
"I have a marvel comics event, need a page for this"

Promt for above page ^^

web/modules/custom/server_ai_content/src/Service/ContentGenerator.php
```diff
@@ -0,0 +1,328 @@
...
+    protected FileSystemInterface $fileSystem,
+  ) {}
+
+  /**
+   * Generate a landing page from a user prompt.
+   *
+   * @param string $prompt
+   *   The user's text prompt.
+   * @param string $contentType
+   *   The node bundle to create.
```
andreytroeglazov: **Security: Paragraph type from AI response not validated**

The `type` from the AI JSON response is passed directly to `paragraph->create()` without validation against `ParagraphSchemaDiscovery::ALLOWED_TYPES`. The allow-list only controls what's described in the prompt, but doesn't enforce at entity creation time. A hallucinating model or prompt injection could cause creation of arbitrary paragraph types (e.g., types requiring entity references, which would fail or create broken entities). Consider validating `$type` against the allowed list before creating.

web/modules/custom/server_ai_content/src/Service/ParagraphSchemaDiscovery.php
```diff
@@ -0,0 +1,251 @@
...
+          if (!empty($subFields)) {
+            $schema[$bundle]['sub_paragraph'] = [
+              'type' => $subBundle,
+              'label' => $bundleInfo[$subBundle]['label'] ?? $subBundle,
+              'field_name' => $field['field_id'],
+              'fields' => $subFields,
+            ];
+          }
+        }
+      }
```
andreytroeglazov: **Bug: Sub-paragraph loop overwrites previous entries**

The inner `foreach ($subBundles as $subBundle)` loop overwrites `$schema[$bundle]['sub_paragraph']` on each iteration. If a paragraph reference field targets multiple sub-bundle types, only the last one survives. Should either collect all sub-bundles into an array, or (if only one is expected) break after the first match.

web/modules/custom/server_ai_content/src/Service/ContentGenerator.php
```diff
@@ -0,0 +1,328 @@
...
+   *
+   * @param array $paragraphData
+   *   Paragraph data with 'type' and 'fields' keys.
+   * @param array $compoundMapping
+   *   Mapping of compound types from ParagraphSchemaDiscovery.
+   *
+   * @return \Drupal\paragraphs\ParagraphInterface|null
+   *   The created paragraph, or NULL on failure.
+   */
+  protected function createParagraph(array $paragraphData, array $compoundMapping): ?ParagraphInterface {
```
andreytroeglazov: **Logic: Hardcoded text_long field detection**

`field_body` and `field_description` are hardcoded to detect text_long fields that need `format`. This is fragile — if a new paragraph type has a differently-named text_long field, it won't get the format set. Consider using the schema metadata (field type from `ParagraphSchemaDiscovery`) to determine if a field is `text_long` rather than matching field names.

web/modules/custom/server_ai_content/tests/src/ExistingSite/AiContentGenerateFormTest.php
```diff
@@ -0,0 +1,38 @@
...
+  public function testAnonymousAccessDenied(): void {
+    $this->drupalGet('/admin/content/ai-generate');
+    $this->assertSession()->statusCodeEquals(Response::HTTP_FORBIDDEN);
+  }
+
+  /**
+   * Test that admin users can access the form.
+   */
+  public function testAdminCanAccessForm(): void {
+    $user = $this->createUser();
```
andreytroeglazov: **Test: Admin role bypasses permission check**

This test gives the user the `administrator` role, which typically has `is_admin: true` and bypasses all permission checks. This means the test doesn't actually verify that the custom `generate ai content` permission works. Use `$this->createUser(['generate ai content'])` instead to properly test the permission.

web/modules/custom/server_ai_content/src/Form/AiContentGenerateForm.php
```diff
@@ -0,0 +1,88 @@
...
+   * {@inheritdoc}
+   */
+  public static function create(ContainerInterface $container): static {
+    $instance = new static();
+    $instance->contentGenerator = $container->get('server_ai_content.content_generator');
+    return $instance;
+  }
+
+  /**
+   * {@inheritdoc}
```
andreytroeglazov: **Code style: Form DI via property assignment instead of constructor**

The `create()` method uses `new static()` + direct property assignment. Drupal coding standards prefer constructor injection with `parent::__construct()`. Consider:
```php
public function __construct(
  protected ContentGenerator $contentGenerator,
) {}
```

web/modules/custom/server_ai_content/src/Service/ContentGenerator.php
```diff
@@ -0,0 +1,328 @@
...
+    $paragraph = $this->entityTypeManager->getStorage('paragraph')->create($values);
+    $paragraph->save();
+
+    return $paragraph;
+  }
+
+  /**
+   * Map a single field value to a Drupal-compatible format.
+   *
+   * @param string $fieldName
```
andreytroeglazov: **Minor: Image filename collision risk**

`'ai-generated-' . time() . '.png'` will collide if multiple images are generated within the same second. Consider using `\Drupal::service('uuid')->generate()` or `uniqid('', TRUE)` for uniqueness.

web/modules/custom/server_ai_content/tests/src/ExistingSite/ContentGeneratorTest.php
```diff
@@ -0,0 +1,94 @@
...
+
+    // Third paragraph is accordion with 2 items.
+    $this->assertEquals('accordion', $paragraphs[2]->bundle());
+    $accordionItems = $paragraphs[2]->get('field_accordion_items')->referencedEntities();
+    $this->assertCount(2, $accordionItems);
+    $this->assertEquals('Question 1', $accordionItems[0]->get('field_title')->value);
+    $this->assertEquals('Question 2', $accordionItems[1]->get('field_title')->value);
+
+    // Clean up.
+    $node->delete();
```
andreytroeglazov: **Test cleanup: Paragraphs not cleaned up**

The test manually calls `$node->delete()` but doesn't clean up the created paragraph entities. If paragraph deletion doesn't cascade from the node (depends on config), orphaned paragraphs will remain. Consider using `$this->markEntityForCleanup($node)` instead, and also marking paragraphs for cleanup, or verifying cascade behavior.

andreytroeglazov:
`paragraphs_ai` is not used anymore, it's done mostly for site building, even paragraphs schema lookup returns yaml which needs parsing and converting to a promt so I gave up on this and removed, as a result no external core dependencies except drupal/ai.

web/modules/custom/server_ai_content/src/Service/ContentGenerator.php
```diff
@@ -0,0 +1,328 @@
...
+          'target_id' => $paragraph->id(),
+          'target_revision_id' => $paragraph->getRevisionId(),
+        ];
+      }
+    }
+
+    /** @var \Drupal\node\NodeInterface $node */
+    $node = $this->entityTypeManager->getStorage('node')->create([
+      'type' => $contentType,
+      'title' => $data['title'] ?? 'AI Generated Page',
```
amitaibu:
```suggestion
      'title' => $data['title'],
```

web/modules/custom/server_ai_content/src/Service/ContentGenerator.php
```diff
@@ -0,0 +1,328 @@
...
+   *   The created unpublished node.
+   */
+  public function createFromParsedData(array $data, string $contentType): NodeInterface {
+    $compoundMapping = $this->schemaDiscovery->getCompoundTypeMapping($contentType);
+    $paragraphEntities = [];
+
+    foreach ($data['paragraphs'] ?? [] as $paragraphData) {
+      $paragraph = $this->createParagraph($paragraphData, $compoundMapping);
+      if ($paragraph) {
+        $paragraphEntities[] = [
```
amitaibu:
```suggestion
        $paragraph_entities[] = [
```

web/modules/custom/server_ai_content/src/Form/AiContentGenerateForm.php
```diff
@@ -0,0 +1,88 @@
...
+   */
+  public function submitForm(array &$form, FormStateInterface $form_state): void {
+    $prompt = $form_state->getValue('prompt');
+
+    try {
+      $node = $this->contentGenerator->generate($prompt, 'landing_page');
```
amitaibu: As a follow-up, we should consider removing the hardcoding of the landing page to allow the creation of other content types, such as News.

web/modules/custom/server_ai_content/src/Service/ContentGenerator.php
```diff
@@ -0,0 +1,328 @@
...
+   *   The node bundle to create.
+   *
+   * @return \Drupal\node\NodeInterface
+   *   The created unpublished node.
+   *
+   * @throws \RuntimeException
+   *   If the AI response cannot be parsed.
+   */
+  public function generate(string $prompt, string $contentType = 'landing_page'): NodeInterface {
+    $schemaDescription = $this->schemaDiscovery->buildPromptDescription($contentType);
```
amitaibu: Here and other variables should be in `snake_case`

```suggestion
    $schema_description = $this->schemaDiscovery->buildPromptDescription($contentType);
```

andreytroeglazov: yes, I will check all, some ihp influence :slightly_smiling_face:

web/modules/custom/server_ai_content/tests/src/ExistingSite/ContentGeneratorTest.php
```diff
@@ -0,0 +1,94 @@
...
+  /**
+   * Test creating a landing page from parsed AI response data.
+   */
+  public function testCreateFromParsedData(): void {
+    /** @var \Drupal\server_ai_content\Service\ContentGenerator $generator */
+    $generator = \Drupal::service('server_ai_content.content_generator');
+
+    $data = [
+      'title' => 'Test AI Generated Page',
+      'paragraphs' => [
```
amitaibu: - Let's create new paragraph types for the test, as module would live in contrib
- We should also test `sub_paragraph`
- Does it know how to create a Media item, or not yet?

web/modules/custom/server_ai_content/src/Form/AiContentGenerateForm.php
andreytroeglazov: It's done for any bundle, we just need to pass it there from form

web/modules/custom/server_ai_content/tests/src/ExistingSite/ContentGeneratorTest.php
andreytroeglazov: > Does it know how to create a Media item, or not yet?

yes, in generateImage in ContentGenerator

web/modules/custom/server_ai_content/src/Form/AiContentGenerateForm.php
amitaibu: We could introduce a new setting for each bundle, allowing an admin to input information about that bundle through the UI. For instance, if we have a Microsite (i.e., OG group), we could specify for AI that any other node created should reference it. This would naturally follow, especially considering that we might enable creating multiple nodes from a single prompt.

amitaibu:
- Once issues here are fixed, let's move code to https://www.drupal.org/project/ai_content_assistant (so this PR will not be merged)
- I'll open follow up issues on d.o.

web/modules/custom/server_ai_content/src/Service/ContentGenerator.php
andreytroeglazov: fixed

web/modules/custom/server_ai_content/src/Service/ParagraphSchemaDiscovery.php
andreytroeglazov: fixed

web/modules/custom/server_ai_content/tests/src/ExistingSite/AiContentGenerateFormTest.php
andreytroeglazov: fixed

web/modules/custom/server_ai_content/src/Service/ContentGenerator.php
andreytroeglazov: fixed

web/modules/custom/server_ai_content/src/Form/AiContentGenerateForm.php
andreytroeglazov: fixed

web/modules/custom/server_ai_content/src/Service/ContentGenerator.php
andreytroeglazov: fixed

web/modules/custom/server_ai_content/tests/src/ExistingSite/ContentGeneratorTest.php
andreytroeglazov: fixed

web/modules/custom/server_ai_content/src/Form/AiContentGenerateForm.php
```diff
@@ -0,0 +1,102 @@
...
+
+    return $form;
+  }
+
+  /**
+   * {@inheritdoc}
+   */
+  public function submitForm(array &$form, FormStateInterface $form_state): void {
+    $prompt = $form_state->getValue('prompt');
+    $content_type = $form_state->getValue('node_type');
```
andreytroeglazov: **Security: No node type creation access check.**

The route only requires `generate ai content` permission, but never checks `create {type} content`. A user with only the AI permission could generate nodes of any content type, bypassing Drupal's per-bundle access control.

Consider adding `_entity_create_access: node` to the route requirements, or checking `$this->entityTypeManager->getAccessControlHandler('node')->createAccess($content_type)` in `buildForm`/`submitForm`.

web/modules/custom/server_ai_content/src/Service/ContentGenerator.php
```diff
@@ -0,0 +1,377 @@
...
+    // Handle image fields — check for image_prompt key.
+    if (is_array($field_value) && isset($field_value['image_prompt'])) {
+      $media = $this->generateImage($field_value['image_prompt']);
+      return $media ? ['target_id' => $media->id()] : NULL;
+    }
+
+    // Handle entity reference fields — single {"target_id": 123}
+    // or multiple [{"target_id": 1}, {"target_id": 2}].
+    if (is_array($field_value) && isset($field_value['target_id'])) {
+      return ['target_id' => (int) $field_value['target_id']];
```
andreytroeglazov: **Security: Entity reference IDs not validated.**

The `target_id` from the AI response is accepted without verifying the entity exists or the user has access. An AI hallucination could reference a non-existent or access-restricted entity. Consider loading the entity and checking `$entity->access('view')` before creating the reference, or at minimum verifying it exists.

web/modules/custom/server_ai_content/src/Form/AiContentGenerateForm.php
```diff
@@ -0,0 +1,102 @@
...
+
+    try {
+      $node = $this->contentGenerator->generate($prompt, $content_type);
+      $this->messenger()->addStatus($this->t('"@title" has been generated as a draft.', [
+        '@title' => $node->getTitle(),
+      ]));
+      $form_state->setRedirectUrl($node->toUrl());
+    }
+    catch (\Throwable $e) {
+      $this->messenger()->addError($this->t('Content generation failed: @message', [
```
andreytroeglazov: **Security: Information disclosure in error message.**

`$e->getMessage()` may expose sensitive internal details (API keys, provider errors, internal paths) to end users. Show a generic user-facing message and only log the full error:

```php
$this->messenger()->addError($this->t('Content generation failed. Please try again or contact an administrator.'));
$this->getLogger('server_ai_content')->error('AI content generation failed: @message', ['@message' => $e->getMessage()]);
```

web/modules/custom/server_ai_content/src/Service/ContentGenerator.php
```diff
@@ -0,0 +1,377 @@
...
+    // Handle link fields.
+    if (is_array($field_value) && isset($field_value['uri'])) {
+      return $field_value;
+    }
+
+    // Handle text_long/text_with_summary fields — set value and format.
+    if (is_string($field_value) && $this->isFormattedTextField($paragraph_type, $field_name)) {
+      return [
+        'value' => $field_value,
+        'format' => 'full_html',
```
andreytroeglazov: **Security: AI-generated content stored with `full_html` format.**

If `full_html` allows `<script>` or `<iframe>` tags (common for admin-level formats), a prompt injection or malformed AI response could introduce XSS when the content is rendered. Consider using a more restrictive text format (e.g., `basic_html`) or sanitizing the AI output before storing.

web/modules/custom/server_ai_content/src/Service/ContentGenerator.php
```diff
@@ -0,0 +1,377 @@
...
+    if (is_array($field_value) && isset($field_value['target_id'])) {
+      return ['target_id' => (int) $field_value['target_id']];
+    }
+    if (is_array($field_value) && isset($field_value[0]['target_id'])) {
+      return array_map(fn($item) => ['target_id' => (int) $item['target_id']], $field_value);
+    }
+
+    // Handle link fields.
+    if (is_array($field_value) && isset($field_value['uri'])) {
+      return $field_value;
```
andreytroeglazov: **Security: Link URI not validated.**

The `uri` from AI output is stored directly without validation. The system prompt instructs the AI to use `route:<nolink>`, but the AI could return any URI scheme. Consider validating the URI scheme (allow only `route:`, `internal:`, `https://`, `http://`) to prevent `javascript:` or other dangerous URI schemes.

web/modules/custom/server_ai_content/src/Service/ContentGenerator.php
```diff
@@ -0,0 +1,377 @@
...
+        ];
+      }
+    }
+
+    /** @var \Drupal\node\NodeInterface $node */
+    $node = $this->entityTypeManager->getStorage('node')->create([
+      'type' => $content_type,
+      'title' => $data['title'],
+      'status' => 0,
+      'field_paragraphs' => $paragraph_entities,
```
andreytroeglazov: **Logic: Hardcoded `field_paragraphs` field name.**

Both here and in `ParagraphSchemaDiscovery::getParagraphBundles()` (line 217), the paragraph field name is hardcoded as `field_paragraphs`. If a content type uses a different paragraph reference field name, this silently produces an empty node with no paragraphs. Consider discovering the paragraph reference field dynamically (by looking for `entity_reference_revisions` fields targeting paragraphs) or making it configurable.

web/modules/custom/server_ai_content/src/Service/ContentGenerator.php
```diff
@@ -0,0 +1,377 @@
...
+   *   Parsed JSON data with 'title' and 'paragraphs' keys.
+   * @param string $content_type
+   *   The node bundle.
+   *
+   * @return \Drupal\node\NodeInterface
+   *   The created unpublished node.
+   */
+  public function createFromParsedData(array $data, string $content_type): NodeInterface {
+    $schema = $this->schemaDiscovery->getSchema($content_type);
+    $compound_mapping = $this->schemaDiscovery->getCompoundTypeMapping($content_type);
```
andreytroeglazov: **Performance: Schema computed twice.**

`getSchema()` is called here, and then `getCompoundTypeMapping()` on the next line calls `getSchema()` again internally. Each call queries entity field definitions and executes entity queries. Consider either passing the schema into `getCompoundTypeMapping()`, or extracting the compound mapping inline from `$schema`.

web/modules/custom/server_ai_content/src/Service/ParagraphSchemaDiscovery.php
```diff
@@ -0,0 +1,413 @@
...
+      $field_type = $definition->getType();
+      $info = [
+        'field_id' => $field_name,
+        'field_label' => (string) $definition->getLabel(),
+        'field_type' => $field_type,
+        'required' => $definition->isRequired(),
+        'cardinality' => $definition->getFieldStorageDefinition()->getCardinality(),
+      ];
+
+      if (in_array($field_type, ['entity_reference', 'entity_reference_revisions'])) {
```
andreytroeglazov: **Style: Missing strict comparison in `in_array()`.**

Drupal coding standards require the third `TRUE` parameter for strict comparison:
```php
if (in_array($field_type, ['entity_reference', 'entity_reference_revisions'], TRUE)) {
```

web/modules/custom/server_ai_content/src/Service/ContentGenerator.php
```diff
@@ -0,0 +1,377 @@
...
+   * Generate an image via DALL-E and create a Media entity.
+   *
+   * @param string $image_prompt
+   *   Description of the image to generate.
+   *
+   * @return \Drupal\media\MediaInterface|null
+   *   The created media entity, or NULL on failure.
+   */
+  protected function generateImage(string $image_prompt): ?MediaInterface {
+    return $this->callDallE($image_prompt);
```
andreytroeglazov: **Minor: Unnecessary `generateImage()` wrapper.**

`generateImage()` just delegates to `callDallE()` with no additional logic. Consider removing the indirection and calling `callDallE()` directly from `mapFieldValue()`.

andreytroeglazov:
Added "Generate with AI" button which redirects to promt form
Selection_097
Selection_098

Now it picks all paragraphs, including paragraph with entity reference fields (last paragraph on the screenshot).
screencapture-drupal-starter-ddev-site-4443-sustainable-urban-gardening-growing-green-city-2026-04-06-18_13_30

amitaibu:
> Now it picks all paragraphs, including paragraph with entity reference fields

How does AI know which content to reference?

Let's address the issues here, and then move dev to d.o. (but not create a release yet)
So this/ new PR will change  to `ddev composer require` and a bunch of config

andreytroeglazov:
> How does AI know which content to reference?

We send titles + ids to it, so it picks relevant to promt.

amitaibu:
> We send titles + ids to it

Link to the code, please. Are we sending _all_ titles/ IDs?

web/modules/custom/server_ai_content/src/Service/ContentGenerator.php
```diff
@@ -0,0 +1,377 @@
...
+  /**
+   * Call DALL-E to generate an image and create a Media entity.
+   *
+   * @param string $image_prompt
+   *   Description of the image to generate.
+   *
+   * @return \Drupal\media\MediaInterface|null
+   *   The created media entity, or NULL on failure.
+   */
+  protected function callDallE(string $image_prompt): ?MediaInterface {
```
amitaibu: We're relying on AI module now, so we don't know for sure consumer is going to use OpenAI, so name should change here

amitaibu:
> Link to the code, please. Are we sending all titles/ IDs?

Found it

https://github.com/Gizra/drupal-starter/pull/1018/changes#diff-570568ad92386c50e86f5628955cf345ef21f68a490def60e3dcd33bfb8e2ab2R343


I have some ideas on how to prevent it from exploding if we get too many items, but that could be a follow up

web/modules/custom/server_ai_content/src/Service/ContentGenerator.php
andreytroeglazov: fixed

andreytroeglazov:
> I have some ideas on how to prevent it from exploding if we get too many items, but that could be a follow up
> Are we sending all titles/ IDs?

50 only for now, I haven't thought how to get relevant list for AI yet

andreytroeglazov:
Updated to alpha release

amitaibu (approved):
Thanks, screenshots and MWG

andreytroeglazov:
image
image


screencapture-drupal-starter-ddev-site-4443-world-batman-2026-04-16-14_19_56

PR #1019, opened 2026-04-02
Use ObjectType::isSuperTypeOf() in CacheableDependency PHPStan rule

bboro:
Replace `method_exists`/`isInstanceOf` check with `ObjectType::isSuperTypeOf()->no()` to avoid false positives for interface types whose concrete implementations implement `CacheableDependencyInterface`.


amitaibu (approved):
Thanks, MWG

PR #1022, opened 2026-04-14
Handle API failures gracefully in release notes

mariano-dagostino:
## Summary
- Wrap GitHub API calls for PR details in try-catch to prevent exceptions from wrong PR/issue numbers from blocking entire release note generation
- Previously, any invalid PR number referenced in a commit message would throw an exception and prevent generating the release notes

## Testing
- Verify release notes can be generated even with commits containing invalid issue/PR references

🦺 Generated with [Opencode](https://github.com/anomalyco/opencode)


AronNovak (approved):
👍

PR #1023, opened 2026-04-17
Meaningful dependabot

AronNovak:
Dependabot is useful, but not when it mass generates PRs for theme compilation dependencies.


amitaibu (approved):

amitaibu:
Thanks

PR #1030, opened 2026-04-27
Use npm ci

balagan73:
Use the read-only npm ci instead of npm install, which can modify package-lock.json file.


AronNovak (approved):
Makes perfect sense!

PR #1032, opened 2026-05-03
Fix Webform external libraries missing from composer.lock

balagan73:
## Summary

- A recent Dependabot update caused Webform external libraries to be dropped from `composer.lock`, triggering warnings on the Drupal status page about missing Webform libraries
- Ran `ddev composer update drupal/webform "drupal/webform-*" --with-dependencies` to regenerate `composer.lock` with the correct Webform library entries

## Test plan

- [x] Visit `admin/reports/status` and confirm no warnings about missing Webform external libraries

### DIFF
--- a/.github/workflows/dependabot-webform.yml
+++ b/.github/workflows/dependabot-webform.yml
@@ -1,0 +1,34 @@
+name: Update Webform libraries after Dependabot
+
+on:
+  pull_request:
+    paths:
+      - 'composer.json'
+      - 'composer.lock'
+
+jobs:
+  update-webform-libraries:
+    if: github.actor == 'dependabot[bot]'
+    runs-on: ubuntu-latest
+    permissions:
+      contents: write
+    steps:
+      - uses: actions/checkout@v4
+        with:
+          ref: ${{ github.head_ref }}
+          token: ${{ secrets.GITHUB_TOKEN }}
+
+      - name: Setup PHP
+        uses: shivammathur/setup-php@v2
+        with:
+          php-version: '8.3'
+
+      - name: Update Webform libraries in composer.lock
+        run: composer update drupal/webform "drupal/webform-*" --with-dependencies --no-interaction
+
+      - name: Commit updated composer.lock
+        run: |
+          git config user.name "github-actions[bot]"
+          git config user.email "github-actions[bot]@users.noreply.github.com"
+          git add composer.lock
+          git diff --staged --quiet || git commit -m "Update Webform libraries in composer.lock" && git push
--- a/.github/workflows/dependabot-webform.yml
+++ b/.github/workflows/dependabot-webform.yml
@@ -31,4 +31,4 @@ jobs:
           git config user.name "github-actions[bot]"
           git config user.email "github-actions[bot]@users.noreply.github.com"
           git add composer.lock
-          git diff --staged --quiet || git commit -m "Update Webform libraries in composer.lock" && git push
+          git diff --staged --quiet || (git commit -m "Update Webform libraries in composer.lock" && git push)
--- a/composer.lock
+++ b/composer.lock
@@ -282,6 +282,21 @@
             "time": "2024-07-16T11:13:48+00:00"
         },
         {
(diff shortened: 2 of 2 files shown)


AronNovak:
> A recent Dependabot update caused Webform external libraries to be dropped from composer.lock, triggering warnings on the Drupal status page about missing Webform libraries

Is it a known issue of Dependabot? Can you please check?
Or is it a configuration issue that we might mitigate via https://docs.github.com/en/code-security/concepts/supply-chain-security/about-the-dependabot-yml-file ?

.github/workflows/dependabot-webform.yml
```diff
@@ -0,0 +1,34 @@
+name: Update Webform libraries after Dependabot
+
+on:
+  pull_request:
+    paths:
+      - 'composer.json'
+      - 'composer.lock'
+
+jobs:
+  update-webform-libraries:
```
AronNovak, line 10: It seems to be strange to add a webform-specific solution, as other libraries / modules can be affected too.

balagan73: Do you mean I should rename the job?

AronNovak: I think the solution is too webform-specific. We should understand the problem and have a fix that works for webform and covers all the similar cases too.

balagan73: Assumption was that dependabot runs without the --plugins or --plugin flag, thus not invoking the composer-merge-plugin, which would pull in the libraries defined by the composer.json that comes with the webform module.

AronNovak: I see, but can we instruct Dependabot to use that flag?

balagan73: It seems we cannot, dependabot runs in a strict environment. We could add the required libraries explicitly to composer.json, keeping or removing the merge-plugin. That way we have to maintain the main composer.json manually when webform updates its libraries.

balagan73: Ok, it seems the issue is not related to the --plugin flag.

> The actual reason webform libraries are missing: It's a chicken-and-egg problem specific to wikimedia/composer-merge-plugin. The merge config points to web/modules/contrib/webform/composer.libraries.json, which lives inside the contrib directory — which is not committed to git. When Dependabot clones the repo, that file doesn't exist yet. When it then runs composer install (to reproduce the current lock state), the merge-plugin can't find the file to merge, so the libraries never enter the lock file.

So the solution is webform specific, as only webform has its libraries listed under the merge-plugin section:

image

## The drawbacks of this solution:

- Extra commit on every Dependabot composer PR (resets checks, clutters history)
- Lost on @dependabot rebase, has to re-run
- Risk of unintended webform version bumps on unrelated PRs
- Slow — full Composer install from scratch each run
- Tightly scoped to webform — any other module with the same pattern needs a duplicated workflow
- Silent failure mode if the update step fails


## without the workflow, the typical reactive flow becomes:

 Dependabot PR merges to main with broken composer.lock
Main is now broken (deploys fail, fresh installs fail)
Someone notices (CI on main, deploy failure, or worst case a developer hitting it locally)
They locally run composer install then composer update drupal/webform --with-dependencies
Open a fix PR with the updated lock file
Get it reviewed and merged
So per Dependabot composer PR, you incur: a broken-main window + a follow-up PR + review cycle. Multiplied by however many Dependabot composer PRs you merge per week.

That's strictly worse than the workflow's drawbacks. The workflow's "extra commit on each PR" trades off against "broken main + extra PR" — the workflow wins on both correctness (main is never broken) and effort (no follow-up PR needed).

The alternative I mentioned (CI check that fails Dependabot PRs missing libraries) shifts the manual work to before merge — you'd push a fix commit to the Dependabot branch instead of opening a separate fix PR after merge. That's only marginally less work than just keeping the auto-fix workflow.

Bottom line: keep the workflow. Its drawbacks are real but mild compared to the alternatives.

AronNovak (approved):
Let's try this then

PR #1033, opened 2026-05-03
Migrate button and line-separator ThemeTraits to SDC

balagan73:
## Why

PEVB rendering previously followed:

```
PEVB Plugin → ThemeTrait (#theme hook) → hook_theme() registration → Twig in theme/templates/
```

This scattered templates across the theme with no explicit props contract and no co-location. Every PEVB plugin required jumping between the module (plugin + ThemeTrait) and the theme (template + `hook_theme()` registration).

SDC (Single Directory Components) is stable in Drupal core since 10.3/11 and is the community-endorsed direction for component-based theming. It improves the pattern by:
- Making props/slots explicit via `.component.yml`
- Co-locating the template with the ThemeTrait — no more theme-switching when building a PEVB plugin
- Eliminating `hook_theme()` registrations

## What changed

- Migrated `LineSeparatorThemeTrait` and `ButtonThemeTrait` from `#theme` to `#type => 'component'`
- Created `components/button/` and `components/line-separator/` directories in `server_general` with `.component.yml` and `.twig` files
- Removed the `server_theme_button` and `server_theme_line_separator` entries from `hook_theme()` in `server_theme.theme`
- Added `@source` to the Tailwind CSS entry point so `.twig` files in `modules/custom/` are picked up by JIT (one-time project setup)
- `ButtonThemeTrait`: URL objects are resolved to strings with `toString(TRUE)` and the resulting cache metadata is bubbled explicitly via `#cache`, since SDC props accept scalars not URL objects

## Note

This branch was created on top of the branch that fixes missing Webform external libraries (which broke `ddev composer install`). Those commits are included here; they should be disregarded as part of this PR's review.


amitaibu:
Thanks, but I believe we don't need the components or SDC since we already have ThemeTraits, which offer type safety unlike the components.

balagan73:
All right. I just ran a few prompts to discover how pevb fits into drupal render pipeline, and Gemini supposed we are already using SDC, as the logical next step. Since I always felt changing to the theme directory to find the relevant templates a nuisance, I thought I give it a try.

Issue #783, opened 2025-01-10
Paragraphs CRUD on own page


amitaibu:
One of the issues with Paragraphs is that its widget can easily get out of hand, piling up with too many entities and sub-entities. You probably know those pages - add, edit, re-order on in a single form

![image](https://github.com/user-attachments/assets/473f23ea-6145-4eba-a12f-efa64db9d1da)

While working on IHP, we've ended up with a simpler, one action per entity == one page
So very rough and un-themed we have the Landing page, where you can create new Paragraphs and edit/ delete existing ones
But the Create and Edit are on a new page

![image (1)](https://github.com/user-attachments/assets/ccc5dcd9-825f-44b1-b94f-1dd11f5a8825)

So again un-themed, and no breadcrumbs - but here you can see the Create page for the Quote paragraph

![image (2)](https://github.com/user-attachments/assets/9cf8c43b-07eb-40f4-93bb-d24e6612bb6b)

Here, a little nicer, on an internal IHP we have - you can see the re-order is also separated out, to its own page

| Click to re-order | Takes you to a new page |
| -- | -- |
| ![image (4)](https://github.com/user-attachments/assets/f1a8ac59-d3fa-4170-9874-2d0b3a70570f) | ![image (5)](https://github.com/user-attachments/assets/b07b7df0-55e8-4e4c-9785-7d827e2aa3db) |

So what I have in mind for a new contrib module:

1. Create a new field widget for Paragraph, where you can click on create new Paragraph types; See existing ones; Re-order link
1. Add those new routes, so Paragraphs are created/ edited/ deleted on own page (we already have the edit and delete with [Paragraphs Edit](https://www.drupal.org/project/paragraphs_edit/))

tanmayk:
Here is rough idea of the design of the widget. We can use Edit, Delete routes provided by `paragraphs_edit`. For adding new, we can have dropdown links which will take us to new page.

<img width="797" height="429" alt="Image" src="https://github.com/user-attachments/assets/166b41c4-3d19-49b1-ad3d-97a55f646ff5" />

To persist the data from other fields while creatng/editing the node, we need to look into [Autosave Form](https://www.drupal.org/project/autosave_form).

Module name can be: `paragraphs_navigator` or `paragraphs_flow`. Took help from AI for name suggestions.

TB: 2h

amitaibu:
Bump to total 6h, to get initial code going

tanmayk:
[...]

PR #1037, opened 2026-05-12
New widget to access paragraph_edit routes directly from the edit form

mariano-dagostino:
#783 

@amitaibu The creation part of this feature is quite complicated, I vote for adding a simple enhancement for now. Maybe combined with #1018 we won't need the creation part at all.

https://github.com/user-attachments/assets/f2de7fbf-60c6-40d1-8e48-e4dbbe37e25c

### DIFF
--- a/web/modules/custom/server_general/src/Plugin/Field/FieldWidget/ParagraphsEditWidget.php
+++ b/web/modules/custom/server_general/src/Plugin/Field/FieldWidget/ParagraphsEditWidget.php
@@ -1,0 +1,90 @@
+<?php
+
+declare(strict_types=1);
+
+namespace Drupal\server_general\Plugin\Field\FieldWidget;
+
+use Drupal\Core\Field\Attribute\FieldWidget;
+use Drupal\Core\Field\FieldItemListInterface;
+use Drupal\Core\Form\FormStateInterface;
+use Drupal\Core\StringTranslation\TranslatableMarkup;
+use Drupal\Core\Url;
+use Drupal\paragraphs\Plugin\Field\FieldWidget\ParagraphsWidget;
+
+/**
+ * Plugin implementation of the 'paragraphs_edit' paragraphs widget.
+ *
+ * Extends the core ParagraphsWidget to add an "Edit standalone" button
+ * that opens the paragraph edit form in a separate page.
+ *
+ * @FieldWidget(
+ *   id = "paragraphs_edit",
+ *   label = @Translation("Paragraphs (edit in new tab)"),
+ *   description = @Translation("Paragraphs widget with edit in new tab option."),
+ *   field_types = {
+ *     "entity_reference_revisions"
+ *   }
+ * )
+ */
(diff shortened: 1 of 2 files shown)


amitaibu:
Thanks. Maybe the existing "Edit" should be replaced by the "Edit Standalone"?

mariano-dagostino:
@amitaibu Done. Also contributed a patch to paragraph_edit to have Cancel button https://git.drupalcode.org/project/paragraphs_edit/-/merge_requests/25


https://github.com/user-attachments/assets/d7ef0d9e-eb40-42fd-9a47-24ec17c0f6c5

The legacy edit inline now lives inside the dropdown.

Screenshot_2026-06-05_17-04-38

amitaibu:
Thanks. To confirm, when Creating a new node, the `Edit` widget won't appear, correct?

mariano-dagostino:
@amitaibu yes, on `node/add` both edit links are not displayed,

Screenshot_2026-06-08_09-48-46

web/modules/custom/server_general/src/Plugin/Field/FieldWidget/ParagraphsEditWidget.php
```diff
@@ -0,0 +1,97 @@
...
+  /**
+   * {@inheritdoc}
+   */
+  public function formElement(FieldItemListInterface $items, $delta, array $element, array &$form, FormStateInterface $form_state): array {
+    $element = parent::formElement($items, $delta, $element, $form, $form_state);
+
+    $field_name = $this->fieldDefinition->getName();
+    $parents = $element['#field_parents'];
+
+    $widget_state = static::getWidgetState($parents, $field_name, $form_state);
```
amitaibu, line 46: Why `static` and not `parent`

mariano-dagostino: @amitaibu I think this is correct, If someone extends the widget and overrides `getWidgetState`, their override would be ignored.  

Below some LLM sumary:

Using static::getWidgetState is correct because:

1. ParagraphsWidget is not final - It can be extended (as we're doing with our ParagraphsEditWidget)
2. WidgetBase is also not final - The entire inheritance chain is open for extension
3. Late static binding is the Drupal pattern - The parent class uses static:: throughout (11 occurrences)

amitaibu: You have a few lines above `$element = parent::formElement` -- so I'm wondering what the diff between the calls is?

mariano-dagostino: I'm not sure to follow your last comment. `static` implies `parent` if the called method is not defined in the current class, so I think is safer to use.

amitaibu (approved):
Thanks

PR #1041, opened 2026-05-14
Install and enable m4032404 module.

### LINKED ISSUE #1040: Install m4032404 module
Install the module to show 404 page instead of 403 pages for unpublished nodes.

balagan73:
Closes #1040 

- Installed and enabled m4032404 module.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -55,6 +55,7 @@
         "drupal/inline_entity_form": "^3.0@RC",
         "drupal/intl_date": "^1.1",
         "drupal/lang_dropdown": "^2.1",
+        "drupal/m4032404": "^2.0",
         "drupal/maxlength": "^3.1",
         "drupal/media_entity_browser": "^3.0",
         "drupal/metatag": "^2.0",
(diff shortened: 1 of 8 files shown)


balagan73:
image
An unpublished page displays 404 for anon user.

AronNovak:
@balagan73 Will it work with Fast404?

balagan73:
@AronNovak Good catch. Actually the Fast404 module mentions m4032404 in its example settings:
https://git.drupalcode.org/project/fast_404/-/blob/8.x-2.x/example.settings.fast404.php?ref_type=heads#L84
In the light of this I think we should use 
` $settings['fast404_not_found_exception'] = TRUE;
`
 so served 404 pages would be coherent, otherwise serving different themed pages work against our intent of hiding which paths exist.

balagan73:
With the setting updated unpublished nodes show unthemed 404 page:
image

AronNovak:
@balagan73 Before it is merged, what's the rationale of this? To prevent ID enumeration? So security or something else?

balagan73:
@AronNovak I need confirmation from @amitaibu, but I think security, so anon users could not tell apart existing paths and non existing paths.

isapisa:
@AronNovak @balagan73 actually, it's a result of SEO. We want 403 to be 404 so they're not indexed and therefore don't show up in search results

PR #1052, opened 2026-06-04
Aggregate nginx logs across all Pantheon app containers

AronNovak:
Pantheon environments can run on multiple application containers, each keeping its own nginx logs. `getLogFile()` only fetched logs from the single container returned by `terminus connection:info --field=sftp_command`, so the DDoS / access-log analysis saw only a fraction of the traffic.

Now resolve every container IP behind the appserver hostname and aggregate their `nginx-access.log` into one file before analysis.

Ref: https://docs.pantheon.io/guides/logs-pantheon/automate-log-downloads

### DIFF
--- a/robo-components/SecurityTrait.php
+++ b/robo-components/SecurityTrait.php
@@ -196,16 +196,84 @@ trait SecurityTrait {
    */
   protected function getLogFile(string $env) {
     $pantheon_info = $this->getPantheonNameAndEnv();
-    $this->_exec('$(terminus connection:info --field=sftp_command ' . $pantheon_info['name'] . ".$env) <<EOF
+    $site_env = $pantheon_info['name'] . ".$env";
+
+    // A Pantheon environment can run on several application containers, each
+    // keeping its own nginx logs. The SFTP command from Terminus only targets a
+    // single container, so we resolve every container IP and aggregate their
+    // logs - otherwise the analysis only sees a fraction of the traffic.
+    // @see https://docs.pantheon.io/guides/logs-pantheon/automate-log-downloads
+    $sftp_username = trim($this->getTerminusConnectionField($site_env, 'sftp_username'));
+    $sftp_host = trim($this->getTerminusConnectionField($site_env, 'sftp_host'));
+    $sftp_port = trim($this->getTerminusConnectionField($site_env, 'sftp_port')) ?: '2222';
+
+    if (empty($sftp_username) || empty($sftp_host)) {
+      throw new \Exception("Could not get SFTP connection information for $site_env");
+    }
+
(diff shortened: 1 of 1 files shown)


robo-components/SecurityTrait.php
```diff
@@ -196,16 +196,84 @@ public function securityCheckDdos(string $env = 'live', int $limit = 25, ?string
    */
   protected function getLogFile(string $env) {
     $pantheon_info = $this->getPantheonNameAndEnv();
-    $this->_exec('$(terminus connection:info --field=sftp_command ' . $pantheon_info['name'] . ".$env) <<EOF
+    $site_env = $pantheon_info['name'] . ".$env";
```
mariano-dagostino, line 199: @AronNovak What about making the `getLogFile` a public method so it can be used to download logs outside the scope of checking bots?

robo-components/SecurityTrait.php
```diff
@@ -196,16 +196,84 @@ public function securityCheckDdos(string $env = 'live', int $limit = 25, ?string
...
+
+    // A Pantheon environment can run on several application containers, each
+    // keeping its own nginx logs. The SFTP command from Terminus only targets a
+    // single container, so we resolve every container IP and aggregate their
+    // logs - otherwise the analysis only sees a fraction of the traffic.
+    // @see https://docs.pantheon.io/guides/logs-pantheon/automate-log-downloads
+    $sftp_username = trim($this->getTerminusConnectionField($site_env, 'sftp_username'));
+    $sftp_host = trim($this->getTerminusConnectionField($site_env, 'sftp_host'));
+    $sftp_port = trim($this->getTerminusConnectionField($site_env, 'sftp_port')) ?: '2222';
```
mariano-dagostino, line 208: ```terminus connection:info site.env --format=json --fields=sftp_username,sftp_port,sftp_host```

returns the 3 values we need as json format. in a single call.

robo-components/SecurityTrait.php
```diff
@@ -196,16 +196,84 @@ public function securityCheckDdos(string $env = 'live', int $limit = 25, ?string
...
+   *
+   * @param string $site_env
+   *   The Pantheon site and environment, e.g. `myproject.live`.
+   * @param string $field
+   *   The connection field to read, e.g. `sftp_host`.
+   *
+   * @return string
+   *   The field value, or an empty string when it could not be retrieved.
+   */
+  protected function getTerminusConnectionField(string $site_env, string $field): string {
```
mariano-dagostino, line 267: Not needed in my opinion, see the `--fields` param above.

mariano-dagostino (changes requested):
@AronNovak oh good find! Looks good, but I have some changes to propose.

PR #1059, opened 2026-07-02
Stale update status cache causes false "not secure" warnings during Pantheon deploy checks

### LINKED ISSUE #1058: Stale update status cache causes false "not secure" warnings during Pantheon deploy checks [1h]
## Problem

`deployPantheonSync()` runs `deployCheckRequirementErrors()` (`robo-components/DeploymentTrait.php` ~L563) as the very last step of deployment, after `env:deploy`, `updb`, `cim`, `cr`, and `deploy:hook` have all completed. That method runs `terminus remote:drush ... -- rq --format=json` (`DeploymentTrait.php:594`).

## Root cause

Drush's `rq` is a CLI invocation with no Drupal route match. Drupal core's `UpdateManager::getProjects()` (`web/core/modules/update/src/UpdateManager.php:148`) delegates to `projectStorage()` (`UpdateManager.php:175-196`), which only clears/rebuilds the `update_project_projects` keyvalue cache (collection `update`) when the current request matches specific admin routes (`update.status`, `system.modules_list`, etc). A drush CLI call never matches those routes, so it always falls back to returning the stale cached snapshot from whenever an admin last visited `/admin/reports/updates` (or no cache at all).

Traced end-to-end via this project's own vendored Drupal core:
`hook_runtime_requirements` → `update_get_available()` → `UpdateManager::getProjects()` → `projectStorage()`.

## Effect

Deploy notifications can falsely report "Not secure! (x.y.z available)" for a module/core version that was in fact the one that was just deployed, because the version comparison used stale pre-deploy cached data instead of the just-deployed code state.

## Proposed fix

In `deployCheckRequirementErrors()` (`DeploymentTrait.php`), before the `rq` call at line 594, add a step to clear the stale cache:

```
terminus remote:drush $pantheon_terminus_environment -- php:eval "\Drupal::keyValueExpirable('update')->delete('update_project_projects');"
```

This forces `update_requirements()` to recompute project/version info from the just-deployed code instead of a pre-deploy snapshot.

balagan73:
Closes #1058

## Summary
- `deployCheckRequirementErrors()` runs `drush rq` as the last step of a Pantheon deploy, but Drush CLI calls never match a Drupal route, so `UpdateManager::projectStorage()` always serves its stale cached `update_project_projects` data instead of recomputing it against the code just deployed.
- This can produce false "Not secure!" warnings for a module/core version that was in fact the one just deployed.
- Fix: clear the `update_project_projects` key in the `update` keyvalue-expirable store via `terminus remote:drush ... -- php:eval` in its own exec stack, immediately before the `rq` check, so the comparison is always computed fresh.

## Test plan

- [x] `ddev phpstan` — no errors
- [x] Verified the exact eval command locally with `ddev drush`: populated the `update_project_projects` key via `Drupal::service('update.manager')->getProjects()`, ran the delete command, confirmed the key was removed from `keyValueExpirable('update')`
- [ ] Not yet verified end-to-end against a live Pantheon deploy (requires terminus auth not available in this environment)

TB: 0.7/1h

### DIFF
--- a/robo-components/DeploymentTrait.php
+++ b/robo-components/DeploymentTrait.php
@@ -588,6 +588,18 @@ trait DeploymentTrait {
   public function deployCheckRequirementErrors(string $environment): void {
     $pantheon_info = $this->getPantheonNameAndEnv();
     $pantheon_terminus_environment = $pantheon_info['name'] . '.' . $environment;
+
+    // Drush CLI calls have no Drupal route match, so
+    // UpdateManager::projectStorage() always returns its stale cached
+    // project data instead of recomputing it, which can produce false
+    // "Not secure!" warnings for the version we just deployed. Clear it
+    // before checking requirements, in its own stack so its output does
+    // not get mixed into the rq JSON response below.
+    $this->taskExecStack()
+      ->stopOnFail()
+      ->exec("terminus remote:drush $pantheon_terminus_environment -- php:eval \"\\Drupal::keyValueExpirable('update')->delete('update_project_projects');\"")
+      ->run();
+
     $task = $this->taskExecStack()
       ->stopOnFail();
     $output = $task


AronNovak (approved):

amitaibu:
Can you please show a screenshot of how it looks before and after

balagan73:
This is what we have now. In another project when we did the update, we got a false positive:
image

PR #1061, opened 2026-07-02
Post live deployment notifications to the "Create/Cut release and deploy" issue instead of the commit-derived issue

### LINKED ISSUE #1060: Post live deployment notifications to the "Create/Cut release and deploy" issue instead of the commit-derived issue [1h]
## Problem

Currently `deployNotify()` in `robo-components/DeploymentTrait.php` (~line 790) determines which GitHub issue to comment on by parsing the deployment commit message (merge-commit or squash-PR patterns) to find an issue number. For live deployments this is the wrong target — live deploys are triggered from a tag push whose commit message doesn't correspond to the originating feature issue. Instead, when deploying to the `live` Pantheon environment, we want to find the latest open issue titled exactly "Create release and deploy" or "Cut release and deploy" (case-insensitive) and post the deployment comment there.

## Agreed design

1. Add a new protected method `deployFindLatestReleaseIssue(string $github_token): ?int` in `DeploymentTrait.php`. It calls `GET /repos/{project}/issues?state=open&sort=created&direction=desc&per_page=100` via the existing `curl` + `taskExec` pattern already used elsewhere in the file, skips entries that have a `pull_request` key (the issues endpoint also returns PRs), and returns the first issue whose title (trimmed, lowercased) exactly equals "create release and deploy" or "cut release and deploy". Returns `NULL` if none found.
2. In `deployNotify()`, when `$pantheon_environment === 'live'`, call this new method first. If it returns an issue number, use it as the sole entry in `$issue_numbers` and skip the commit-message parsing entirely. If it returns `NULL`, fall through unchanged to the existing commit-message/PR-based logic.
3. QA and test environments are untouched, keeping today's commit-message-based behavior.
4. Out of scope: a separate pre-existing bug where `deployNotify()` reads `getenv('GITHUB_COMMIT_MESSAGE')` but only `TRAVIS_COMMIT_MESSAGE` is ever exported by `ci-scripts/prepare_deploy.sh` — leave that as-is, do not fix here.

No changes needed to `.github/workflows/deploy.yml` or `ci-scripts/*.sh`; this is a self-contained change to `robo-components/DeploymentTrait.php`.

balagan73:
Closes #1060

## Summary
- Live deploys are triggered from a release tag push, so the deployment commit message doesn't point back to the originating feature issue — `deployNotify()` would fail to find (or find the wrong) issue to comment on.
- Added `deployFindLatestReleaseIssue()`, which looks up the latest open GitHub issue whose title starts with "Create release" or "Cut release" (case-insensitive, tolerant of trailing text like "and deploy", "& deploy", or a `[2h]` timebox suffix).
- `deployNotify()` now uses this lookup first when `$pantheon_environment === 'live'`; if no matching issue is found, it falls back unchanged to the existing commit-message/PR-based logic. QA and test environments are unaffected.

## Test plan
- [x] `ddev phpcs robo-components/DeploymentTrait.php` — no violations
- [x] `ddev phpstan analyse robo-components/DeploymentTrait.php` — no errors
- [x] Verified `deployFindLatestReleaseIssue()` against real repos: returns correct issue number
- [x] Ran `deployNotify('live')` end-to-end in a temporary dry-run (print instead of POST) against another repo with deployment issue, confirming it targets real issue and composes the expected comment body

TB: 1/1h

### DIFF
--- a/robo-components/DeploymentTrait.php
+++ b/robo-components/DeploymentTrait.php
@@ -780,6 +780,51 @@ trait DeploymentTrait {
   }
 
   /**
+   * Finds the latest open release issue to notify about live deployments.
+   *
+   * @param string $github_token
+   *   The GitHub API token.
+   *
+   * @return int|null
+   *   The number of the latest matching open issue, or NULL if no open
+   *   issue titled starting with "Create release" or "Cut release" was
+   *   found. The rest of the title varies (e.g. "and deploy", "& deploy",
+   *   a trailing timebox like "[2h]"), so only the leading words are
+   *   matched.
+   */
+  protected function deployFindLatestReleaseIssue(string $github_token): ?int {
+    $release_issue_title_prefixes = [
+      'create release',
+      'cut release',
+    ];
+
+    $result = $this->taskExec("curl -H \"Authorization: token $github_token\" \"https://api.github.com/repos/" . self::$githubProject . "/issues?state=open&sort=created&direction=desc&per_page=100\"")
+      ->printOutput(FALSE)
+      ->run()
+      ->getMessage();
+
+    $issues = json_decode($result);
+    if (!is_array($issues)) {
+      return NULL;
+    }
+
+    foreach ($issues as $issue) {
+      // The issues endpoint also returns pull requests, skip those.
+      if (isset($issue->pull_request)) {
+        continue;
+      }
+      $title = strtolower(trim($issue->title));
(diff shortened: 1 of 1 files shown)


AronNovak (commented):
Looks good, the usual way for such improvements is to test-drive at a client project, if it turns to be useful and robust, we can port it to DS. The `latestreleaseissue` seems to be fragile for me. It would be great to somehow robustly link an issue.

PR #1062, opened 2026-07-14
Fix extra whitespace in class attributes generated by Twig templates

wombatbuddy:
Before

https://github.com/user-attachments/assets/9e8ae628-4490-44fb-938c-e625f773080e

After

https://github.com/user-attachments/assets/982b4288-5920-4d7d-adba-340ce5546ab1

## Description

Several Twig templates generate `class` attributes with extra whitespace: leading spaces, multiple consecutive spaces, and/or trailing spaces.

This PR fixes the affected templates so that rendered `class` attributes contain only single spaces between class names, with no leading or trailing whitespace.

## Steps to Reproduce

1. Open the `/style-guide` page.
2. Open Chrome DevTools and switch to the **Console** tab.
3. Run the following script to scan the rendered HTML for `class` attributes with extra whitespace:

```js
fetch(location.href)
  .then(r => r.text())
  .then(html => {
    const pattern = /class="(?:\s[^"]*|[^"]*\s{2,}[^"]*|[^"]*\s)"/g;
    const matches = [...html.matchAll(pattern)].map(m => m[0]);
    console.table(matches);
    console.log('Class attributes with extra whitespace found:', matches.length);
  });
```

4. Observe that the script reports multiple matches — `class` attributes containing a leading space, multiple consecutive spaces, or a trailing space.

## Changes Made

Fixed 14 Twig templates across the theme that generated class attributes with leading, trailing, or duplicate whitespace. See the "Files changed" tab for the full list.

Added a PHPUnit test that renders the /style-guide page and fails if any class attribute contains leading, trailing, or duplicate whitespace.

## How to Verify

1. Open `/style-guide` after this change.
2. Re-run the script above in the console.
3. Confirm `Matches found: 0`.

## Checklist

- [ ] Verified with the console script above that no `class` attributes with extra whitespace remain on `/style-guide`
- [ ] Visually checked affected components to confirm no styling regressions

### DIFF
--- a/web/modules/custom/server_general/tests/src/ExistingSite/ServerGeneralRedundantSpacingTest.php
+++ b/web/modules/custom/server_general/tests/src/ExistingSite/ServerGeneralRedundantSpacingTest.php
@@ -16,9 +16,33 @@ use Symfony\Component\HttpFoundation\Response;
  * over raw markup, so that quoting style, comments, and non-attribute text
  * don't produce false positives/negatives.
  *
+ * The check is scoped to an allow-list of root selectors that correspond to
+ * markup owned by this project (custom theme + custom modules), so that it
+ * doesn't flag markup provided by Drupal Core or contrib modules (e.g. the
+ * admin toolbar, contextual links, or the messages region) that we can't fix
+ * ourselves.
+ *
  * @group server_general
  */
 class ServerGeneralRedundantSpacingTest extends ServerGeneralTestBase {
+
+  /**
+   * Root selectors that scope the test to markup owned by this project.
+   *
+   * Each selector is checked both for its own class attribute and for the
+   * class attribute of every descendant element.
+   *
+   * @var string[]
+   */
+  private const SCANNED_ROOT_SELECTORS = [
+    // Style-guide component sandbox (everything rendered by
+    // StyleGuideController).
+    'dl.accordion',
+    // Main navigation menu (menu--main.html.twig), covers both the
+    // mobile <ul class="main-menu"> and desktop <div class="main-menu">
+    // variants.
+    '.main-menu',
+  ];
 
   /**
    * Tests that no class attribute contains redundant whitespace.
@@ -33,13 +57,25 @@ class ServerGeneralRedundantSpacingTest extends ServerGeneralTestBase {
     $this->assertSession()->statusCodeEquals(Response::HTTP_OK);
 
     $html = (string) $this->getSession()->getPage()->getContent();
-
     $crawler = new Crawler($html);
-    $classValues = $crawler->filter('[class]')->extract(['class']);
+
+    $classValues = [];
+    foreach (self::SCANNED_ROOT_SELECTORS as $selector) {
+      // The root element's own class attribute.
+      $classValues = array_merge(
+        $classValues,
+        $crawler->filter($selector)->extract(['class'])
+      );
+      // The class attribute of every descendant element.
+      $classValues = array_merge(
+        $classValues,
+        $crawler->filter($selector . ' [class]')->extract(['class'])
+      );
+    }
 
     $this->assertNotEmpty(
       $classValues,
-      'No elements with a class attribute were found on /style-guide — check that the page rendered correctly.'
+      'No elements with a class attribute were found in the scanned regions (' . implode(', ', self::SCANNED_ROOT_SELECTORS) . ') on /style-guide — check that the page rendered correctly or that the selectors still match the markup.'
     );
 
     $offenders = [];
--- a/web/modules/custom/server_general/tests/src/ExistingSite/ServerGeneralRedundantSpacingTest.php
+++ b/web/modules/custom/server_general/tests/src/ExistingSite/ServerGeneralRedundantSpacingTest.php
@@ -1,0 +1,66 @@
+<?php
+
+declare(strict_types=1);
+
+namespace Drupal\Tests\server_general\ExistingSite;
+
+use Symfony\Component\DomCrawler\Crawler;
+use Symfony\Component\HttpFoundation\Response;
+
+/**
+ * Checks that class attributes on the /style-guide have no redundant spaces.
+ *
+ * Checks that class attributes on the /style-guide page have no redundant
+ * leading, trailing, or repeated spaces. Uses the DOM crawler
+ * (proper HTML parsing) to extract class attribute values, rather than regex
+ * over raw markup, so that quoting style, comments, and non-attribute text
+ * don't produce false positives/negatives.
+ *
+ * @group server_general
+ */
+class ServerGeneralRedundantSpacingTest extends ServerGeneralTestBase {
+
+  /**
+   * Tests that no class attribute contains redundant whitespace.
+   */
+  public function testNoRedundantSpacesInClassAttributes(): void {
+    // Log in as admin to get access to the style-guide page.
+    $user = $this->createUser();
+    $user->addRole('administrator');
+    $user->save();
+    $this->drupalLogin($user);
+    $this->drupalGet('/style-guide');
+    $this->assertSession()->statusCodeEquals(Response::HTTP_OK);
+
+    $html = (string) $this->getSession()->getPage()->getContent();
+
+    $crawler = new Crawler($html);
+    $classValues = $crawler->filter('[class]')->extract(['class']);
+
+    $this->assertNotEmpty(
+      $classValues,
+      'No elements with a class attribute were found on /style-guide — check that the page rendered correctly.'
+    );
+
+    $offenders = [];
+    foreach ($classValues as $index => $classValue) {
+      // Leading space, trailing space, or 2+ consecutive spaces.
+      if (preg_match('/^\s|\s{2,}|\s$/', $classValue) === 1) {
+        $offenders[$index] = $classValue;
+      }
+    }
+
+    $this->assertEmpty(
+      $offenders,
+      sprintf(
+        "Found %d class attribute(s) with redundant spaces on /style-guide:\n%s",
+        count($offenders),
+        implode("\n", array_map(
+          static fn (string $v): string => '"' . $v . '"',
+          $offenders
+        ))
+      )
+    );
+  }
+
+}
--- a/web/themes/custom/server_theme/templates/menu--main.html.twig
+++ b/web/themes/custom/server_theme/templates/menu--main.html.twig
@@ -66,9 +66,9 @@
       <li{{ item.attributes.addClass(item_classes) }}>
         {% set item_tag = item.below ? "button" : "a" %}
         {% if menu_level == 0 %}
-          <{{ item_tag }} href="{{ item.url }}" class="menu-root-item pb-2 flex items-center text-xl md:text-2xl w-full text-left{{ item.in_active_trail ? ' active' }}">{{ item.title }}{% if item.below %} <span class="expand-indicator flex justify-center items-center w-6 h-6 border-1 rounded-full border-white ml-4 {{ item.in_active_trail ? 'expand-indicator--expanded' }}"></span>{% endif %}</{{ item_tag }}>
+          <{{ item_tag }} href="{{ item.url }}" class="menu-root-item pb-2 flex items-center text-xl md:text-2xl w-full text-left{{ item.in_active_trail ? ' active' }}">{{ item.title }}{% if item.below %} <span class="expand-indicator flex justify-center items-center w-6 h-6 border-1 rounded-full border-white ml-4{{ item.in_active_trail ? ' expand-indicator--expanded' }}"></span>{% endif %}</{{ item_tag }}>
         {% elseif menu_level == 1 %}
-          <{{ item_tag }} href="{{ item.url }}" class="menu-sub-item pb-2 flex items-center text-lg md:text-xl text-left{{ item.in_active_trail ? ' active' }}">{{ item.title }} {% if item.below %} <span class="expand-indicator flex justify-center items-center w-6 h-6 border-1 rounded-full border-white ml-4 {{ item.in_active_trail ? 'expand-indicator--expanded' }}"></span>{% endif %}</{{ item_tag }}>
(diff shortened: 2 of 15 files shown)


web/themes/custom/server_theme/templates/menu--main.html.twig
```diff
@@ -66,11 +66,11 @@
       <li{{ item.attributes.addClass(item_classes) }}>
         {% set item_tag = item.below ? "button" : "a" %}
         {% if menu_level == 0 %}
-          <{{ item_tag }} href="{{ item.url }}" class="menu-root-item pb-2 flex items-center text-xl md:text-2xl w-full text-left {{ item.in_active_trail ? 'active' }}">{{ item.title }}{% if item.below %} <span class="expand-indicator flex justify-center items-center w-6 h-6 border-1 rounded-full border-white ml-4 {{ item.in_active_trail ? 'expand-indicator--expanded' }}"></span>{% endif %}</{{ item_tag }}>
+          <{{ item_tag }} href="{{ item.url }}" class="menu-root-item pb-2 flex items-center text-xl md:text-2xl w-full text-left{{ item.in_active_trail ? ' active' }}">{{ item.title }}{% if item.below %} <span class="expand-indicator flex justify-center items-center w-6 h-6 border-1 rounded-full border-white ml-4 {{ item.in_active_trail ? 'expand-indicator--expanded' }}"></span>{% endif %}</{{ item_tag }}>
```
bboro: Missed out the span

```suggestion
          <{{ item_tag }} href="{{ item.url }}" class="menu-root-item pb-2 flex items-center text-xl md:text-2xl w-full text-left{{ item.in_active_trail ? ' active' }}">{{ item.title }}{% if item.below %} <span class="expand-indicator flex justify-center items-center w-6 h-6 border-1 rounded-full border-white ml-4{{ item.in_active_trail ? ' expand-indicator--expanded' }}"></span>{% endif %}</{{ item_tag }}>
```

web/themes/custom/server_theme/templates/menu--main.html.twig
```diff
@@ -66,11 +66,11 @@
...
         {% elseif menu_level == 1 %}
-          <{{ item_tag }} href="{{ item.url }}" class="menu-sub-item pb-2 flex items-center text-lg md:text-xl text-left {{ item.in_active_trail ? 'active' }}">{{ item.title }} {% if item.below %} <span class="expand-indicator flex justify-center items-center w-6 h-6 border-1 rounded-full border-white ml-4 {{ item.in_active_trail ? 'expand-indicator--expanded' }}"></span>{% endif %}</{{ item_tag }}>
+          <{{ item_tag }} href="{{ item.url }}" class="menu-sub-item pb-2 flex items-center text-lg md:text-xl text-left{{ item.in_active_trail ? ' active' }}">{{ item.title }} {% if item.below %} <span class="expand-indicator flex justify-center items-center w-6 h-6 border-1 rounded-full border-white ml-4 {{ item.in_active_trail ? 'expand-indicator--expanded' }}"></span>{% endif %}</{{ item_tag }}>
```
bboro:
```suggestion
          <{{ item_tag }} href="{{ item.url }}" class="menu-sub-item pb-2 flex items-center text-lg md:text-xl text-left{{ item.in_active_trail ? ' active' }}">{{ item.title }} {% if item.below %} <span class="expand-indicator flex justify-center items-center w-6 h-6 border-1 rounded-full border-white ml-4{{ item.in_active_trail ? ' expand-indicator--expanded' }}"></span>{% endif %}</{{ item_tag }}>
```

web/themes/custom/server_theme/templates/menu--main.html.twig
```diff
@@ -158,11 +158,11 @@ We call a macro which calls itself to render the full tree.
             <li{{ item.attributes.setAttributes('class', item_classes) }}>
               {% set item_tag = item.below ? "button" : "a" %}
               {% if menu_starting_level == 0 %}
-                <{{ item_tag }} data-menu-child="{{ item.below ? key|split(':')|last : '' }}" href="{{ item.url }}" class="menu-root-item flex items-center base leading-11 lg:text-xl font-header pb-2 w-full text-left {{ item.in_active_trail ? 'active' }}">{{ item.title }} {% if item.below %}<span class="expand-indicator flex justify-center items-center w-6 h-6 border-1 rounded-full border-white ml-4 {{ item.in_active_trail ? 'expand-indicator--expanded' }}"></span>{% endif %}</{{ item_tag }}>
+                <{{ item_tag }} data-menu-child="{{ item.below ? key|split(':')|last : '' }}" href="{{ item.url }}" class="menu-root-item flex items-center base leading-11 lg:text-xl font-header pb-2 w-full text-left{{ item.in_active_trail ? ' active' }}">{{ item.title }} {% if item.below %}<span class="expand-indicator flex justify-center items-center w-6 h-6 border-1 rounded-full border-white ml-4 {{ item.in_active_trail ? 'expand-indicator--expanded' }}"></span>{% endif %}</{{ item_tag }}>
```
bboro:
```suggestion
                <{{ item_tag }} data-menu-child="{{ item.below ? key|split(':')|last : '' }}" href="{{ item.url }}" class="menu-root-item flex items-center base leading-11 lg:text-xl font-header pb-2 w-full text-left{{ item.in_active_trail ? ' active' }}">{{ item.title }} {% if item.below %}<span class="expand-indicator flex justify-center items-center w-6 h-6 border-1 rounded-full border-white ml-4{{ item.in_active_trail ? ' expand-indicator--expanded' }}"></span>{% endif %}</{{ item_tag }}>
```

web/themes/custom/server_theme/templates/menu--main.html.twig
```diff
@@ -158,11 +158,11 @@ We call a macro which calls itself to render the full tree.
...
               {% elseif menu_starting_level == 1 %}
-                <{{ item_tag }} data-menu-child="{{ item.below ? key|split(':')|last : '' }}" href="{{ item.url }}" class="menu-sub-item flex items-center pb-2 text-left {{ item.in_active_trail ? 'active' }}">{{ item.title }} {% if item.below %}<span class="expand-indicator flex justify-center items-center w-6 h-6 border-1 rounded-full border-white ml-4 {{ item.in_active_trail ? 'expand-indicator--expanded' }}"></span>{% endif %}</{{ item_tag }}>
+                <{{ item_tag }} data-menu-child="{{ item.below ? key|split(':')|last : '' }}" href="{{ item.url }}" class="menu-sub-item flex items-center pb-2 text-left{{ item.in_active_trail ? ' active' }}">{{ item.title }} {% if item.below %}<span class="expand-indicator flex justify-center items-center w-6 h-6 border-1 rounded-full border-white ml-4 {{ item.in_active_trail ? 'expand-indicator--expanded' }}"></span>{% endif %}</{{ item_tag }}>
```
bboro:
```suggestion
                <{{ item_tag }} data-menu-child="{{ item.below ? key|split(':')|last : '' }}" href="{{ item.url }}" class="menu-sub-item flex items-center pb-2 text-left{{ item.in_active_trail ? ' active' }}">{{ item.title }} {% if item.below %}<span class="expand-indicator flex justify-center items-center w-6 h-6 border-1 rounded-full border-white ml-4{{ item.in_active_trail ? ' expand-indicator--expanded' }}"></span>{% endif %}</{{ item_tag }}>
```

web/modules/custom/server_general/tests/src/ExistingSite/ServerGeneralRedundantSpacingTest.php
```diff
@@ -0,0 +1,66 @@
...
+    $user->addRole('administrator');
+    $user->save();
+    $this->drupalLogin($user);
+    $this->drupalGet('/style-guide');
+    $this->assertSession()->statusCodeEquals(Response::HTTP_OK);
+
+    $html = (string) $this->getSession()->getPage()->getContent();
+
+    $crawler = new Crawler($html);
+    $classValues = $crawler->filter('[class]')->extract(['class']);
```
bboro: `$classValues = $crawler->filter('dl.accordion [class]')->extract(['class']);`

This avoids filtering classes provided by Core or any other module, and keeps only the classes produced by our components.

web/themes/custom/server_theme/templates/server-theme-container-narrow.html.twig
```diff
@@ -1,15 +1,14 @@
...
-
-<div class="{{ color_class }} {{ py_class }}">
+<div class="{{ classes }}">
   <div class="container-narrow w-full">
     {{ element }}
   </div>
 </div>
 
 
+
```
bboro: remove extra spaces added

web/modules/custom/server_general/tests/src/ExistingSite/ServerGeneralRedundantSpacingTest.php
wombatbuddy: Good point, but this change will exclude our menu components from the test since they doesn't live inside a dl.accordion wrapper. We need to keep the menu covered to check its spacing.

bboro: Can we target the menu specifically? If so maybe we can do 2 passes, one for dl.accordion and one for the menu? That way we can avoid core items and cover the menu as well.

wombatbuddy: We could do that — one selector for `dl.accordion`, one for the menu:

* `dl.accordion` — the style-guide component sandbox (everything `StyleGuideController` renders).
* `.main-menu` — the root class applied in `menu--main.html.twig` to both the mobile (`<ul>`) and desktop (`<div>`) variants of the main menu. It's specific to the main menu template, so it won't accidentally match other menus, and it's not tied to whichever region/block-layout position the menu block ends up in.

This keeps Core/contrib-generated markup (admin toolbar, contextual links, messages region, etc.) out of scope, same as your original suggestion, while still covering the menu.

One addition on top of a straight 2-pass "one selector per pass": for each root we need to check the class attribute of the root element itself, not just its descendants. A selector like `.main-menu [class]` (descendant combinator) only matches elements inside `.main-menu` — it skips class on `.main-menu` itself.

One caveat: `.main-menu` is a CSS class tied to the theme's naming, not a stable test hook — if it gets renamed during a styling refactor, this test needs updating alongside it, same as `dl.accordion` today. Happy to add a `data-qa-scope` attribute instead if we want the test fully decoupled from styling classes, but this keeps the diff minimal for now.

```php
<?php

declare(strict_types=1);

namespace Drupal\Tests\server_general\ExistingSite;

use Symfony\Component\DomCrawler\Crawler;
use Symfony\Component\HttpFoundation\Response;

/**
 * Checks that class attributes on the /style-guide have no redundant spaces.
 *
 * Checks that class attributes on the /style-guide page have no redundant
 * leading, trailing, or repeated spaces. Uses the DOM crawler
 * (proper HTML parsing) to extract class attribute values, rather than regex
 * over raw markup, so that quoting style, comments, and non-attribute text
 * don't produce false positives/negatives.
 *
 * The check is scoped to an allow-list of root selectors that correspond to
 * markup owned by this project (custom theme + custom modules), so that it
 * doesn't flag markup provided by Drupal Core or contrib modules (e.g. the
 * admin toolbar, contextual links, or the messages region) that we can't fix
 * ourselves.
 *
 * @group server_general
 */
class ServerGeneralRedundantSpacingTest extends ServerGeneralTestBase {

  /**
   * Root selectors that scope the test to markup owned by this project.
   *
   * Each selector is checked both for its own class attribute and for the
   * class attribute of every descendant element.
   *
   * @var string[]
   */
  private const SCANNED_ROOT_SELECTORS = [
    // Style-guide component sandbox (everything rendered by
    // StyleGuideController).
    'dl.accordion',
    // Main navigation menu (menu--main.html.twig), covers both the
    // mobile <ul class="main-menu"> and desktop <div class="main-menu">
    // variants.
    '.main-menu',
  ];

  /**
   * Tests that no class attribute contains redundant whitespace.
   */
  public function testNoRedundantSpacesInClassAttributes(): void {
    // Log in as admin to get access to the style-guide page.
    $user = $this->createUser();
    $user->addRole('administrator');
    $user->save();
    $this->drupalLogin($user);
    $this->drupalGet('/style-guide');
    $this->assertSession()->statusCodeEquals(Response::HTTP_OK);

    $html = (string) $this->getSession()->getPage()->getContent();
    $crawler = new Crawler($html);

    $classValues = [];
    foreach (self::SCANNED_ROOT_SELECTORS as $selector) {
      // The root element's own class attribute.
      $classValues = array_merge(
        $classValues,
        $crawler->filter($selector)->extract(['class'])
      );
      // The class attribute of every descendant element.
      $classValues = array_merge(
        $classValues,
        $crawler->filter($selector . ' [class]')->extract(['class'])
      );
    }

    $this->assertNotEmpty(
      $classValues,
      'No elements with a class attribute were found in the scanned regions (' . implode(', ', self::SCANNED_ROOT_SELECTORS) . ') on /style-guide — check that the page rendered correctly or that the selectors still match the markup.'
    );

    $offenders = [];
    foreach ($classValues as $index => $classValue) {
      // Leading space, trailing space, or 2+ consecutive spaces.
      if (preg_match('/^\s|\s{2,}|\s$/', $classValue) === 1) {
        $offenders[$index] = $classValue;
      }
    }

    $this->assertEmpty(
      $offenders,
      sprintf(
        "Found %d class attribute(s) with redundant spaces on /style-guide:\n%s",
        count($offenders),
        implode("\n", array_map(
          static fn (string $v): string => '"' . $v . '"',
          $offenders
        ))
      )
    );
  }

}
```

bboro: Yeah sounds good. It's fine if the test needs updating when classnames changes.

bboro (approved):

bboro:
Thanks!

PR #1063, opened 2026-07-15
Fix undefined method setAttributes() in menu--main.html.twig

wombatbuddy:
**What**
Replaces the non-existent `Attribute::setAttributes()` call with `Attribute::addClass()` in the `menu_links_desktop` macro of `menu--main.html.twig`.

```diff
- <li{{ item.attributes.setAttributes('class', item_classes) }}>
+ <li{{ item.attributes.addClass(item_classes) }}>
```

**Why `addClass()` and not `setAttribute('class', ...)`**
`addClass()` appends to any classes already present on `item.attributes`, whereas `setAttribute('class', ...)` would overwrite the whole attribute, wiping out anything set by other code paths.

This also brings `menu_links_desktop` in line with the mobile macro `menu_links`, which already uses `addClass()` correctly for the same purpose — the two macros are now consistent.

### DIFF
--- a/web/themes/custom/server_theme/templates/menu--main.html.twig
+++ b/web/themes/custom/server_theme/templates/menu--main.html.twig
@@ -155,7 +155,7 @@ We call a macro which calls itself to render the full tree.
               menu_starting_level != 0 ? 'my-2.5 px-2.5 lg:my-0 lg:px-0',
             ] %}
 
-            <li{{ item.attributes.setAttributes('class', item_classes) }}>
+            <li{{ item.attributes.addClass(item_classes) }}>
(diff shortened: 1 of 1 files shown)


wombatbuddy:
Hi @amitaibu, just a gentle ping on this PR. The Copilot review confirmed the fix for the Twig runtime error in menu--main.html.twig. Whenever you have a moment, could you please take a look or merge it? Thanks.

amitaibu:
Sorry, I thought I'd already marked it for review by @bboro . Did it now :smile:

bboro (approved):
I think the original intent was `setAttribute` (without s at the end 😅). But this works too and is semantically better.

PR #1065, opened 2026-08-07
Extract Robo commands into shared gizra/drupal-robo package

mariano-dagostino:
## What

Extracts the Robo command traits from this repo into the new standalone package **[gizra/drupal-robo](https://github.com/Gizra/drupal-robo)** (tagged `v1.0.0`), so every project pulls fixes and improvements from one place with `composer update gizra/drupal-robo`.

The package is pulled via a Composer **VCS repository** (no Packagist), so there is nothing to publish — a later move to Packagist would just drop the `repositories` entry.

## Changes here

- **composer.json** — add the `vcs` repository and `gizra/drupal-robo: ^1.0` to `require-dev`.
- **RoboFile.php** — compose `ProjectConfigTrait` and supply the three project-specific values (`getThemeName()`, `getGithubProject()`, `getInstalledLanguages()`); the `INSTALLED_LANGUAGES` constant is gone.
- **robo-components/** — the 10 shared trait files now live in the package and are deleted here. `BootstrapTrait` stays (it scaffolds *new* projects from this template) and is updated to use `getThemeBasePath()` / `getDeploySyncExcludes()`.

`RoboComponents\` autoloads from both the package (`src/`) and the local `robo-components/` (BootstrapTrait) — Composer merges PSR-4 paths for the same namespace.

## How it was parameterized

Project-specific values that used to be hardcoded static properties now come from `ProjectConfigTrait` getters. Three are required (theme name, github project, installed languages); the rest (`getThemeBasePath()`, `getAdminUser()`, `getPoFilesPath()`, `getDeploySyncExcludes()`) ship with defaults and are only overridden when a project diverges.

## Verification

- `ddev robo list` — all commands enumerate (project `bootstrap:project` + package `theme:compile`, `deploy:*`, `locale:*`, `pantheon:*`, `security:*`, `update:modules`, `phpcs`, …), no fatals.
- `phpcs` on `RoboFile.php` + `robo-components/BootstrapTrait.php` — clean.
- `phpstan` on `RoboFile.php` + `robo-components/` — no errors.
- Package CI (in the package repo) runs phpcs + phpstan + phpunit on PHP 8.3/8.4.

## Follow-up (not in this PR)

- `ddev phpcs` no longer sniffs `robo-components/` (the package dropped that path). BootstrapTrait was linted manually here; a `getPhpcsPaths()` getter could restore automatic coverage if wanted.


AronNovak (approved):
It's a pending improvement for a while, nice! 👍 How we would do per-project overrides? Can we just have a trait in the derived project to override some parts?

PR #1068, opened 2026-08-14
Fix autoloader mismatch on deploy

iswariya:
## Summary
- Add a fixed `autoloader-suffix` in `composer.json` to prevent Composer from generating a random suffix on each install, which caused file content changes without size changes.
- Add `--checksum` (`-c`) flag to the rsync command in `DeploymentTrait.php` so that files whose content changed but whose size stayed the same are properly synced to the Pantheon artifact repo.

## Test plan
- [ ] Verify `composer install` generates autoloader files with the `DrupalStarter` suffix
- [ ] Verify deployment to Pantheon correctly syncs autoloader files even when file sizes don't change

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -139,6 +139,7 @@
     },
     "config": {
         "sort-packages": true,
+        "autoloader-suffix": "DrupalStarter",
         "allow-plugins": {
             "composer/installers": true,
             "cweagans/composer-patches": true,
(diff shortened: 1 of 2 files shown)


composer.json
```diff
@@ -139,6 +139,7 @@
     },
     "config": {
         "sort-packages": true,
+        "autoloader-suffix": "DrupalStarter",
```
amitaibu, line 142: Does this need to change, though, from project to project?

iswariya: @amitaibu It's safer to keep it project specific. 
@mariano-dagostino Please provide your thoughts on this?

mariano-dagostino: The issue was, because we don't pin autoloader-suffix, composer derives it from composer.lock's content hash; So any specific value here will work, it does not need to be project specific.

mariano-dagostino (approved):

PR #1074, opened 2026-08-28
Take the News teaser and search label from the content type

AronNovak:
`NodeNews::buildFull` renders the content type label, but the teaser cards, the featured teaser and the search result build theirs from a hardcoded `t('News')`. On a site that renames the News content type, or runs in a language where that string is translated, the two disagree — the full node shows the content type label while the cards still read "News". The label now comes from the content type in all four view modes; the style guide keeps passing a literal, like it already does for the search result.

### DIFF
--- a/web/modules/custom/server_general/tests/src/ExistingSite/ServerGeneralNodeNewsTest.php
+++ b/web/modules/custom/server_general/tests/src/ExistingSite/ServerGeneralNodeNewsTest.php
@@ -2,6 +2,8 @@
 
 namespace Drupal\Tests\server_general\ExistingSite;
 
+use Drupal\node\Entity\NodeType;
+use Drupal\node\NodeInterface;
 use Symfony\Component\HttpFoundation\Response;
 
 /**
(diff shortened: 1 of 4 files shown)


AronNovak:
The bug happened at https://petoovodaert.hu/aktualitasok, which is a Hungarian-only site. At the node page, it correctly said the Hungarian equivalent of News, as the node type was translated, but at the listing paragraph, it remained in English, both places were translatable, but two sources from the very same thing.

amitaibu (commented):
Thanks. Can we add a test?

AronNovak:
Added `ServerGeneralNodeNewsTest::testContentTypeLabel`: it renames the News content type and asserts the new label shows up in all four view modes (full, teaser, featured, search_index). Verified it fails on the teaser against the old code.

The test also uncovered a crash in `buildFeatured`, which passed `NULL` as the image for a news node without a featured image while the trait requires an array — fixed to `[]`, like `buildTeaser` already did.

web/modules/custom/server_general/tests/src/ExistingSite/ServerGeneralNodeNewsTest.php
```diff
@@ -67,4 +69,54 @@ public function testOgMetatags() {
...
+      'moderation_state' => 'published',
+    ]);
+
+    $node_type = NodeType::load('news');
+    $original_label = $node_type->label();
+    $label = 'Bulletin ' . $this->randomMachineName();
+
+    $node_type->set('name', $label)->save();
+
+    try {
```
amitaibu, line 89: Can we remove the `try`?

amitaibu (approved):
Thanks

PR #1076, opened 2026-08-28
#1070 Add AI-crawler-aware robots.txt baseline

mariano-dagostino:
Closes #1070.

## What
Replaces the stock Drupal scaffold `robots.txt` with an explicit, commented, per-vendor baseline for AI crawlers.

- **Names each bot family separately** — OpenAI (`GPTBot`/`OAI-SearchBot`/`ChatGPT-User`), Anthropic (`ClaudeBot`/`Claude-SearchBot`/`Claude-User`), Perplexity (`PerplexityBot`/`Perplexity-User`), Meta (`Meta-ExternalAgent`) — so the policy is tunable per bot, per project.
- **Default: allow** all named AI bots full access to original content (retrieval, index, training).
- **Opts out of `Google-Extended`** (`Disallow: /`), the free Gemini/Vertex training opt-out that doesn't affect Search or AI Overviews.
- **README "AI crawlers"** documents the path-scoped pattern for restricting a training bot to part of the site, and bot-identity verification (IP range / rDNS, best at the CDN/WAF layer).
- **Test** — `ServerGeneralRobotsTxtTest` asserts every bot is named and Google-Extended is disallowed.

The `robots.txt` scaffold override is already pinned to `false` in `composer.json`, so this file is managed by the project and won't be overwritten.

## Follow-up
The CDN/WAF IP/rDNS *enforcement* (issue item 4) as an optional PHP snippet lands in a separate PR, since it's infra-flavored and may not belong in the app layer.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

Issue #1070, opened 2026-08-27
GEO: Add an AI-crawler-aware robots.txt baseline [2h]


isapisa:
### Issue

`robots.txt` in the starter is still the unmodified Drupal scaffold, with no rules distinguishing AI crawlers by vendor or purpose. Each vendor runs independently-controlled bots for training, search-indexing, and live-retrieval — a blanket "AI bot" rule can't express "allow retrieval, block training," so every new client build currently inherits this gap and has to solve it from scratch.

### To Do

- Ship an explicit, commented `robots.txt` baseline naming each AI bot family separately (OpenAI's `GPTBot`/`OAI-SearchBot`/`ChatGPT-User`, Anthropic's `ClaudeBot`/`Claude-SearchBot`/`Claude-User`, Perplexity's `PerplexityBot`/`Perplexity-User`, `Meta-ExternalAgent`), defaulting to allow retrieval/index bots and allow training bots for original content
- Add a `Disallow: Google-Extended` rule (free opt-out from Gemini/Vertex training, no effect on Search or AI Overviews)
- Document the path-scoped pattern for restricting training bots to a specific content type (e.g. licensed/rights-restricted collections), for projects that need it
- Add a CDN/WAF rule that verifies AI-bot traffic against each vendor's published IP range/reverse-DNS before honoring its user-agent string
- Write tests

isapisa:
// cc @amitaibu

mariano-dagostino:
Done as part of #1076 and #1077

PR #1077, opened 2026-08-28
#1070 Add opt-in AI-bot identity verification snippet

mariano-dagostino:
Follow-up to #1076 (robots.txt baseline), closing out #1070's CDN/WAF item.

## What
`robots.txt` is advisory — any client can send `User-agent: GPTBot`. This adds `web/sites/ai_bot_verification.php`, an **opt-in** snippet (modeled on `bot_trap_protection.php`) that *enforces* bot identity via **forward-confirmed reverse DNS**:

1. Request claims to be a known AI bot (`GPTBot`, `ClaudeBot`, `PerplexityBot`, `Meta-ExternalAgent`, …).
2. Its source IP must reverse-resolve to the vendor domain (`.openai.com`, `.anthropic.com`, `.perplexity.ai`, `.facebook.com`).
3. That hostname must forward-resolve back to the same IP.
4. Fails any step → `403`. Ordinary traffic (non-AI UA) is untouched and never triggers a DNS lookup.

## Deliberately not wired in
`bot_trap_protection.php` is `require`d from the settings files; this one is **not**. It's shipped as a documented, self-contained opt-in (see the file's docblock for how to enable) because per-request DNS enforcement in PHP is a fallback — a CDN/WAF verified-bot rule (e.g. Cloudflare) is the better home. Left available for projects that want it without a WAF.

Kept to a single file so it doesn't conflict with #1076's README changes; a README pointer can follow once both land.

### DIFF
--- a/web/sites/ai_bot_verification.php
+++ b/web/sites/ai_bot_verification.php
@@ -1,0 +1,99 @@
+<?php
+
+/**
+ * @file
+ * Forward-confirmed reverse DNS verification for AI crawlers.
+ *
+ * robots.txt only asks well-behaved bots to follow a policy, and any client
+ * can send a "User-agent: GPTBot" header. This snippet enforces bot identity:
+ * when a request CLAIMS to be one of the known AI bots, its source IP must
+ * reverse-resolve to the vendor's domain, and that hostname must
+ * forward-resolve back to the same IP (forward-confirmed reverse DNS).
+ * Requests that fail verification are impersonators, and get a 403.
+ *
+ * Requests whose user agent does not match a known AI bot are left untouched,
+ * so ordinary traffic never triggers a DNS lookup — only requests claiming to
+ * be an AI bot pay the cost.
+ *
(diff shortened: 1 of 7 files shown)


web/sites/ai_bot_verification.php
```diff
@@ -0,0 +1,99 @@
...
+
+// Not an AI bot we verify — leave the request alone.
+if ($claimed_suffixes === NULL) {
+  return;
+}
+
+// Step 1: reverse DNS. The hostname must end in a vendor suffix. Prefixing the
+// hostname with a dot forces the suffix to match on a DNS label boundary, so
+// "notopenai.com" cannot pass as ".openai.com".
+$hostname = gethostbyaddr($remote_addr);
```
AronNovak: This can easily eat PHP runners if name resolution is slow for any reasons.

mariano-dagostino: nice catch, now we have a new implementation that pulls the range know of ips and compare against that.

AronNovak (approved):

PR #1082, opened 2026-09-14
phpcs: anchor the ignore patterns

bboro:
Anchors the phpcs `--ignore` patterns so each one matches only a directory of that exact name.

`--ignore` values are regular expressions, so a bare `.parcel-cache` means "any character followed by parcel-cache". Nothing in this repo matches that by accident today, so file coverage is unchanged: 143 files scanned under `modules/custom` both before and after. The point is the next dotted pattern someone adds cannot silently skip unrelated paths. A bare `.ai`, for instance, would match every path holding "Trait", "mail" or "detail", and phpcs reports nothing when it skips a file.

`node_modules` and the theme's `dist` folder stay bare, since neither holds a regex metacharacter.


mariano-dagostino (commented):
@bboro I wonder if we should be more strict on the other names then, like 'node_modules/`

bboro:
We can do `/node_modules/` which will limit it to just folders named `node_modules`. But I think we're ok largely, as mostly it's concerning Regex and `.` character which means `any character` in regex

PR #1081, opened 2026-09-11
Check that new comments stay short in CI

AronNovak:
Adds a CI job that fails when a pull request adds a comment over 30 words (or a Markdown paragraph over 60), measuring only the blocks it touches so existing long comments are left alone. Docblocks count up to their first tag, and core scaffold files, DDEV-generated add-on files and the `AGENTS.md` symlink are skipped. Replayed over the last 20 merges to `main`, it flagged only the long docblocks in recent feature PRs and nothing in dependency bumps.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

https://claude.ai/code/session_01VmzyYYgLMdGD7zZvqFHiyX

PR #1083, opened 2026-09-15
Comment length check: Elm and TypeScript comments are never measured

anvmn:
Follows #1081.

The comment length check now measures Elm and TypeScript files. Before this, it ignored `.elm` and `.ts`, so a project with an Elm client or Playwright tests written in TypeScript had most of its new comments go unmeasured.

- Elm: `--` line comments and `{- -}` block comments. The `|` that opens a doc comment (`{-| ... -}`) is not counted as a word.
- TypeScript: `//` and `/* */`, read the same way as JavaScript, so a docblock counts up to its first tag.


amitaibu:
@anvmn This is a good idea. Let's get it back?

anvmn:
@amitaibu 

Closed it because there's no TypeScript / Elm code on Drupal starter, but we sure can restore it if you think it brings added value.

amitaibu:
Yeah, let's have it - we'll need it in a few projects

amitaibu (approved):
@anvmn can you please confirm it's working well on a project with TS and Elm? If so, MWG

anvmn:
@amitaibu  works correctly on E-Heza's Elm and TypeScript.

